PR #3, opened 2020-05-06
Use assert to check some parameters in quoteTypedBlock

zhangxp1998:


src/main/scala/lms/core/codegen/C_codegen.scala
```diff
@@ -213,15 +213,20 @@ class ExtendedCCodeGen extends CompactScalaCodeGen with ExtendedCodeGen {
...
+          case (s, i) =>
+            val mod = mods(i)
+            if (autoArgType) s"auto$mod ${quote(s)}"
+            else s"${remap(typeMap(s))}$mod ${quote(s)}"
+        }.mkString(", ")
+      case None => b.in.map{s =>
+          if (autoArgType) s"auto ${quote(s)}"
+          else s"${remap(typeMap(s))} ${quote(s)}"
+      }.mkString(", ")
+    }
```
feiwang3311, line 229: seems an improvement IMO.
What do you think @Kraks ?

Kraks: It looks good to me, I think what can be further improved is to somehow avoid duplicating
```
if (autoArgType) s"auto ${quote(s)}"
else s"${remap(typeMap(s))} ${quote(s)}"
```

TiarkRompf:
LGTM, thanks Kelvin

PR #4, opened 2020-05-19
[CodeGen] fix code gen logic for all unit tests in lms_clean

feiwang3311:
The unit tests were broken for a while because our test use strict string comparison, which is sometimes overly strong. However, it might be good to keep the string output (codegen) as is.

This PR `reverts` some of the earlier changes in the code base so that all tests pass now. I am sure about some of them (which are just one more or one less space) but not sure about others. Please review to see if your intended features are unfortunately reverted.

### DIFF
--- a/src/main/scala/lms/core/codegen/scala_codegen.scala
+++ b/src/main/scala/lms/core/codegen/scala_codegen.scala
@@ -186,14 +186,15 @@ class CompactScalaCodeGen extends CompactTraverser {
     }
     withWraper(wraper _)(f)
   }
-  def quoteBlock(b: Block): Unit = {
+  def quoteBlock(b: Block, argType: Boolean = false): Unit = {
     def eff = quoteEff(b.ein)
-    if (b.in.length == 0) {
-      quoteBlock(traverse(b)) //FIXME(GW): discard eff if arg length == 0?
-    } else {
-      val argsStr = b.in.map(arg => quote(arg)).mkString("(", ", ", s")$eff => ")
-      quoteBlock(argsStr)(traverse(b))
-    }
+    val args = b.in.map({ arg =>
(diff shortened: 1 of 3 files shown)


src/main/scala/lms/core/codegen/scala_codegen.scala
```diff
@@ -186,14 +186,15 @@ class CompactScalaCodeGen extends CompactTraverser {
...
-      val argsStr = b.in.map(arg => quote(arg)).mkString("(", ", ", s")$eff => ")
-      quoteBlock(argsStr)(traverse(b))
-    }
+    val args = b.in.map({ arg =>
+      val tp = if (argType) (": Int") else "" //FIXME(GW): get the type annotation, add a typeMap?
+      quote(arg) + tp
+    })
+    val argsStr = if (args.length == 1 && !argType) s"${args.head}$eff => "
+                  else args.mkString("(", ", ", s")$eff => ")
+    quoteBlock(argsStr)(traverse(b))
```
feiwang3311, line 197: @Kraks please check if this change breaks any of your features.

Kraks: It looks good to me, I will have a test on my code and come back later.

TiarkRompf:
Thanks for looking into this @feiwang3311 ! Does this also address the issue Yi raised with the tutorials? I thought there was something more complicated going on

feiwang3311:
> Thanks for looking into this @feiwang3311 ! Does this also address the issue Yi raised with the tutorials? I thought there was something more complicated going on

Not yet. What Yi raised in address in this PR https://github.com/TiarkRompf/lms-clean/pull/5
and we have to update the checked file in this PR: https://github.com/scala-lms/tutorials/pull/12

This is totally my fault. We had the discussion to address this last semester. I never pushed the commit.

PR #5, opened 2020-05-20
[Backend] fix effect for the Shonan tests

feiwang3311:


src/main/scala/lms/core/stub.scala
```diff
@@ -106,6 +106,13 @@ object Adapter extends FrontEnd {
           curEffects.get(as).filter({ case (_, lrs) => lrs contains rs.asInstanceOf[Sym] }).isDefined } =>
         Some(Const(()))
 
+      case ("array_set", List(as: Exp, i, rs)) if ({
```
GSAir, line 109: Can you add a comment that show the optimization as we have above.

// as(i) = rs; ...; as(i) = rs => as(i) = rs; ...; ()   side condition no brite in-between

src/main/scala/lms/core/backend.scala
```diff
@@ -234,9 +234,26 @@ class GraphBuilder {
...
+                      prevSoft ++= deps.sdeps
+                      prevSoft += lastWrite
+                      prevHard ++= deps.hdeps
+                      true
+                  case _ => false
+                })) {
+                  ()
+                } else {
+                  prevHard += latest(lw);
+                }
```
GSAir, line 254: Weird way to write an if statement...

```scala
if (s == "array_set") {
   findDefinition(lastWrite) match {
                  case Some(Node(_, "array_set", List(as2, idx2, value2), deps)) if (as(0) == as2 && as(1) == idx2) =>
                      prevSoft ++= deps.sdeps
                      prevSoft += lastWrite
                      prevHard ++= deps.hdeps
                  case _ =>
                      prevHard += latest(lw);
}
```

I am not convinced this change is valid when `findDefinition` is a Some() and reflect here is false. Have you taken that into account?

To me it seems that `lr (lastWrite)` will be added to `prevSoft`. It may be correct, but what is the use case?

We should have a test case that illustrate that situation.

Also we are starting to add exception in the reflect function, which was against the design decision initially. As mentioned in the comment.

TiarkRompf: Agreed. Let me stress how important it is to keep reflectEffect clean. This function is a core workhorse of the framework, and everybody doing serious work with LMS will need to read it sooner or later to understand how things work internally. Right now it's gotten pretty much impossible to understand all the logic, even for me. 

So we should refactor for clarity. Helper methods like 'getLatentEffect' were an early attempt in that direction. There should be no special cases in the method, and helper methods should serve as proper extension points. Given the central role of this method, it is also really important to keep in mind that this is an extreme performance hot spot.

src/main/scala/lms/core/stub.scala
feiwang3311: will do

PR #7, opened 2020-05-23
[EffectSystem] refactor and cleaning

feiwang3311:


src/main/scala/lms/core/backend.scala
```diff
@@ -205,55 +204,49 @@ class GraphBuilder {
...
-            // gather effect dependencies
-            val prevHard = new mutable.ListBuffer[Sym]
-            val prevSoft = writes.flatMap(e => curEffects.get(e) match {
-              case Some((lw, lr)) => prevHard += latest(lw); if (!reflectHere) lr else Nil
-              case _ => prevHard += latest(e); Nil
-            })
-            if (reifyHere) prevHard += curBlock
-            prevHard ++= reads.map(getLastWrite(_))
-
+            val (prevHard, prevSoft) = gatherEffectDeps(reads, writes, s, as:_*)
```
feiwang3311, line 237: code refactored into `gatherEffectDeps`

src/main/scala/lms/core/stub.scala
```diff
@@ -80,6 +80,38 @@ object Adapter extends FrontEnd {
...
+          // write has hard dependencies on declaration (if declared locally) or block (if declared globally, i.e., out of current block)
+          prevHard += latest(key);
+      })
+      // gather effect dependencies 2): if we enforce reifyHere (reify in current block), add curBlock as hard dependencies
+      if (reifyHere) prevHard += curBlock
+      // gather effect dependencies 3): handle read keys (i.e., reads have hard dependencies on previous write)
+      reads.foreach(key => prevHard += getLastWrite(key))
+      (prevHard.toSet, prevSoft.toSet)
+    }
+
```
feiwang3311, line 114: this function overwrites the `gatherEffectDeps` case and check for finer-grained write keys and deps

src/main/scala/lms/core/stub.scala
```diff
@@ -106,6 +138,13 @@ object Adapter extends FrontEnd {
           curEffects.get(as).filter({ case (_, lrs) => lrs contains rs.asInstanceOf[Sym] }).isDefined } =>
         Some(Const(()))
 
+      // as(i) = rs; ...; as(i) = rs => as(i) = rs; ...; () side condition no write in-between
+      case ("array_set", List(as: Exp, i, rs)) if ({curEffects.get(as).flatMap({ case (lw, _) => findDefinition(lw)}) match {
+          case Some(Node(_, "array_set", List(as2, idx2, value2), _)) if (as == as2 && i == idx2 && value2 == rs) => true
+          case _ => false
+        }
+      }) => Some(Const(()))
+
```
feiwang3311, line 147: a test for this optimization is added in `test_arrays.scala`

src/test/scala/lms/core/test_arrays.scala
```diff
@@ -200,4 +200,16 @@ class ArrayTest extends TutorialFunSuite {
     }
     check("realloc_dead2", driver.code, "c")
   }
+  test("array_opt_set_equal") {
+    val driver = new DslDriverC[Int,Unit] {
+      @virtualize
+      def snippet(arg: Rep[Int]) = {
+        val x = NewArray[Int](10)
+        x(0) = arg
+        x(0) = arg
```
feiwang3311, line 209: this test is for an optimization that removes redundant array writes.

src/main/scala/lms/core/stub.scala
GSAir: I think it is good to have an override, but that would potentially break everything. The whole logic has to be rewritten.
I don't know if the backend should give that much control to the front end.


Can we make it composable somehow? I didn't think of a solution, but giving something like:

given a list of (cond, soft, hard)

cond: (s: Sym, n: Node) => Boolean, condition to apply the transformation:
```
s == "array" && n == Node(_, "array_set", List(as2, idx2, value2), deps)) && (as(0) == as2 && as(1) == idx2)
```
soft: Node => Set[Sym]
hard: Node => Set[Sym]

I don't know if it is generic enough. But in the case the node doesn't match any condition, use the generic one.

src/main/scala/lms/core/backend.scala
```diff
@@ -205,55 +204,49 @@ class GraphBuilder {
...
             // prevSoft --= prevHard
-            val summary = EffectSummary(prevSoft.toSet, prevHard.toSet, reads, writes)
+            val summary = EffectSummary(prevSoft, prevHard, reads, writes)
             val res = reflect(sm, s, as:_*)(summary)
-            for (key <- reads) {
+            // update effect environments (curEffects, curLocalReads, and curLocalWrites)
+            curLocalReads ++= reads
+            curLocalWrites ++= writes
+            reads.foreach { key =>
```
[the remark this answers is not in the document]
feiwang3311: OK. I can change it back.

src/main/scala/lms/core/backend.scala
```diff
@@ -276,6 +269,26 @@ class GraphBuilder {
...
+      case Some((lw, lr)) =>
+        // write has hard dependencies on previous write (this is conservative for array case, Store, Ctrl,...)
+        prevHard += latest(lw);
+        // write has soft dependencies on previous read (just enforcing order, do not enforcing the reads to be scheduled)
+        if (!reflectHere) prevSoft ++= lr else Nil // FIXME(feiw) why not adding soft dependencies when not reflectHere?
+      case _ =>
+        // write has hard dependencies on declaration (if declared locally) or block (if declared globally, i.e., out of current block)
+        prevHard += latest(key);
+    }}
+    // gather effect dependencies 2): if we enforce reifyHere (reify in current block), add curBlock as hard dependencies
```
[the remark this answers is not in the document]
feiwang3311: Thanks a lot @GSAir . Indeed, I do not really understand the reasoning behind `reifyHere` (or the `reflectHere` in this same file). My understanding is that, by adding `curBlock` to the hard deps, it forces the node to be scheduled `within` the world of the `curBlock` (i.e., it cannot be `code-motioned` to out of the block).

Let's investigate a bit more and come up with a more appropriate comment for this piece of code.

src/main/scala/lms/core/stub.scala
feiwang3311: yeah. I think doing `case` with `falling back to the generic case` is much better :) Will push a fix soon

src/main/scala/lms/core/backend.scala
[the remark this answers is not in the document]
GSAir: the "code motion" is not really the point of effects. I know it is part of it, but it is more important to see it as: in order to produce the correct result this statement has to be generated before that statement.

This is also the case with vanilla reify/reflect so that's not the particularity of the *Here methods. I am pretty sure that *Here are just optimization, it should work with the vanilla ones. You said you found code difference without the *Here. Can you show an example? It would explain why it is interesting to have the *Here methods.

src/main/scala/lms/core/stub.scala
```diff
@@ -80,6 +80,15 @@ object Adapter extends FrontEnd {
 
   class MyGraphBuilder extends GraphBuilder {
 
+    override def gatherEffectDepsWrite(s: String, as: Seq[Def], lw: Sym, lr: Seq[Sym]): (Seq[Sym], Seq[Sym]) =
+    findDefinition(latest(lw)) match {
+      case Some(Node(_, "array_set", as2, deps)) if (s == "array_set" && as.init == as2.init) =>
+        // If latest(lw) is setting the same array at the same index, we do not add hard dependence but soft dependence
+        // In addition, we need to inherite the soft and hard deps of latest(lw)
+        (deps.sdeps.toSeq ++ Seq(latest(lw)), deps.hdeps.toSeq)
```
GSAir, line 88: the order of dependencies doesn't matter: latest(lw) +: deps.sdeps.toSeq.
Why not keeping a Set?

feiwang3311: make sense. will fix :)

src/main/scala/lms/core/stub.scala
feiwang3311: fixed

src/main/scala/lms/core/stub.scala
feiwang3311: fixed :)

src/main/scala/lms/core/backend.scala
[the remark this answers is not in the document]
feiwang3311: That difference was a mistake. I turned off `reifyHere` but still used `isPure` to compute the effects of `if` nodes, which is just wrong.... It needs more careful investigation to see what is optimized by `reifyHere`

PR #9, opened 2020-05-24
[EffectSystem] fix effect for closure

feiwang3311:


src/main/scala/lms/core/backend.scala
```diff
@@ -325,15 +325,13 @@ class GraphBuilder {
...
-      for ((x: Exp, xin) <- args zip argsSym) {
-          if (fwrites(xin)) writes += x
-          else if (freads(xin))  reads += x
-      }
+      // For @ we need to replace the effect on parameters to the actual arguments.
+      // the asInstanceOf seems unsafe at first glance. However, it is not a problem since a standalone block
+      // should never be an argument in function application.
+      val updated_fwrites: Set[Exp] = fwrites.map(k => if (argsSym.contains(k)) args(argsSym.indexOf(k)).asInstanceOf[Exp] else k)
+      val updated_freads: Set[Exp] = freads.map(k => if (argsSym.contains(k)) args(argsSym.indexOf(k)).asInstanceOf[Exp] else k)
+      (reads ++ updated_freads, writes ++ updated_fwrites)
```
feiwang3311, line 333: Here we simply replace the effect symbols for parameters to the corresponding arguments. All other effect symbols remain unchanged and are added to the `reads` and `writes`.

src/main/scala/lms/core/backend.scala
```diff
@@ -322,51 +322,88 @@ class GraphBuilder {
...
+      val (ref, wef) = getLatentEffect(x)
+      (r ++ ref, w ++ wef)
+    }
 
-      // For @ we need to transfer effect on parameters
-      // on the actual values
-      for ((x: Exp, xin) <- args zip argsSym) {
-          if (fwrites(xin)) writes += x
-          else if (freads(xin))  reads += x
-      }
```
feiwang3311, line 357: The getLatentEffect function is heavily refactored.

src/test/scala/lms/core/test_lambda.scala
```diff
@@ -98,6 +98,241 @@ class LambdaTest extends TutorialFunSuite {
...
 
+  // test("nameless_lambda") {
+  //   val driver = new DslDriver[Int,Unit] {
+  //     @virtualize
+  //     def snippet(arg: Rep[Int]) = {
+  //       printf("%d\n", (fun[Int, Int] { (n: Rep[Int]) => n + 1 })(arg))
+  //     }
+  //   }
+  //   System.out.println(indent(driver.code))
+  // }
```
feiwang3311, line 109: cannot get this test to work syntactically. Help?

src/test/scala/lms/core/test_lambda.scala
```diff
@@ -98,6 +98,241 @@ class LambdaTest extends TutorialFunSuite {
...
+  //         val g = fun { (m: Rep[Int]) => x = 1; n + m}
+  //         (g, x)
+  //       }
+  //       val ff = f(arg)
+  //       val res = ff._1(arg + 1)
+  //       printf("%d, %d", ff._2, res)
+  //     }
+  //   }
+  //   System.out.println(indent(driver.code))
+  // }
```
feiwang3311, line 334: this test is waiting for Guannan's PR that can support staged Tuple

src/main/scala/lms/core/backend.scala
```diff
@@ -322,51 +322,88 @@ class GraphBuilder {
...
+  // 2. update the lambda latent effects (replace parameters with arguments)
+  // The sub functon also returns the result of the function because the result might
+  //    be another function that is applied (we need to get its latent effects).
+  def getApplyLatentEffect(f: Sym, args: Def*): ((Set[Exp], Set[Exp]), Option[Exp]) = {
+    val (reads, writes) = getLatentEffect(args: _*)
+    val ((freads, fwrites), (preads, pwrites), res) = getLambdaLatentEffect(f)
+    // For @ we need to replace the effect on parameters to the actual arguments.
+    // the asInstanceOf seems unsafe at first glance. However, it is not a problem since a standalone block
+    // should never be an argument in function application.
+    ((reads ++ freads ++ preads.map(args(_).asInstanceOf[Exp]), writes ++ fwrites ++ pwrites.map(args(_).asInstanceOf[Exp])), res)
```
GSAir, line 399: Why did the filter diseapear? Are you making sure that there is no extra effect being returned here?

```
val f = fun { (x: Int) =>
   val arr = NewArray[Int](x)
   x + arr(0)
}
```
We shouldn't carry the fact that there is a read on `arr` from an outside point of view. The STORE effect is still valid though.

```

Comment (feiwang3311) [reply src/main/scala/lms/core/backend.scala:421]: the block effects don't contain effects for local variables. So `arr` will not be an outstanding effect from outside of the block.

File: src/main/scala/lms/core/backend.scala [hunk 6]
```diff
@@ -322,51 +322,88 @@ class GraphBuilder {
...
+  }
+
+  // The main function for getting latent effects
+  // Latent effects are effects in Blocks (conditional, loop, lambda)
```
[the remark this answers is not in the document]
feiwang3311: Actually I am not so sure about the definition of Latent Effect. Initially I thought it should be `effects of functions`, which are latent because they show up in applications rather than function definition. 
However, the code seems to treat blocks of conditionals and loops as `latent effects` too, which is not that accurate IMO since they are not in any way `latent`. 

What do you think a proper description of latent effect is?

PR #10, opened 2020-06-10
Add support for List/Map/Set/Tuple

Kraks:
- This PR adds support for immutable data structures, `List`/`Map`/`Set`/`Tuple`, as well as Scala codegen of them.
- Take `List` as an example, `ListOps` provides basic operations, which should be (at least) mixed with `Base` when used. `ListOpsOpt` is similar to `ListOps`, but additionally provides some optimizations based on partially-static data (not complete though). 
- In my own use, I also have C++ codegen, which uses the `immer` library. But I think it is probably not everyone would need, so they are not included in this PR.
- @feiwang3311 For mutable collections (eg Array), I think we can also refactor and reorganize in this modular way.

### DIFF
--- a/src/main/scala/lms/collection/ListOps.scala
+++ b/src/main/scala/lms/collection/ListOps.scala
@@ -1,0 +1,186 @@
+package lms.collection.immutable
+
+import lms.core._
+import lms.util._
+import lms.core.stub._
+import lms.core.Backend._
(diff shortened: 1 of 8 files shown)


Kraks:
Merge this now, related issus (if any) will be fixed in future PR.

PR #16, opened 2020-06-12
[frontend] simplify frontend

feiwang3311:


src/main/scala/lms/core/frontend.scala
```diff
@@ -73,19 +73,7 @@ class FrontEnd {
...
-        if (b.isPure)
-          INT(g.reflect("@",f,x.x))
-        else {
-          INT(g.reflectEffectSummary("@",f,x.x)(g.getEffKeys(b)))
-        }
-      case _ =>
-        INT(g.reflectWrite("@",f,x.x)(CTRL))
-    }
-  }
+  def APP(f: Exp, x: INT): INT = INT(g.reflect("@", f, x.x))
```
feiwang3311, line 76: this effect analysis appears unnecessary since the backend is handling it already.

PR #17, opened 2020-06-14
[Documentation] initial commit for code walk latex files

feiwang3311:


Kraks:
Thanks for adding more documents! However, I just wonder if using LaTeX is too heavyweight for this, especially everything perhaps will be changing rapidly. I tend to consider inlined documents in .scala is the firs tstep. Markdown is a another lightweight choice, and it is viewable just in browsers. We can also use `pandoc` to convert Markdowns to LaTeXs at some later point when it becomes necessary.

feiwang3311:
> Thanks for adding more documents! However, I just wonder if using LaTeX is too heavyweight for this, especially everything perhaps will be changing rapidly. I tend to consider inlined documents in .scala is the firs tstep. Markdown is a another lightweight choice, and it is viewable just in browsers. We can also use `pandoc` to convert Markdowns to LaTeXs at some later point when it becomes necessary.

make sense. I started with comments, but the comment might be too much if we want to talk about examples. PDF can also be viewed from browser (maybe not in PR, for some reason). There is no fundamental difference between using markdown and latex :) . I am just more familiar with latex syntax.

PR #18, opened 2020-06-14
[Frontend] reorganize stub.scala

feiwang3311:
I spent some time to reorganize the stub.scala file, which was quite messy. 
One thing to note is that I pulled ArrayOps and RangeOps out of Base, which requires the Flare repo to update a bit. I have the Flare change as a separate PR.

### DIFF
--- a/src/test/scala/lms/core/test_lambda.scala
+++ b/src/test/scala/lms/core/test_lambda.scala
@@ -408,4 +408,20 @@ class LambdaTest extends TutorialFunSuite {
       driver.eval(2)
     })
   }
+
+  test("mutual_recursion") {
+    val driver = new DslDriver[Int,Unit] {
+      @virtualize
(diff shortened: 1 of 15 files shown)


src/test/scala/lms/core/test_lambda.scala
```diff
@@ -408,4 +408,20 @@ class LambdaTest extends TutorialFunSuite {
...
+          if (x == 1) true else even(x - 1)
+        }
+        lazy val even: Rep[Int=>Boolean] = fun { (x: Rep[Int]) =>
+          if (x == 0) true else odd(x - 1)
+        }
+        printf("%d", even(a))
+      }
+    }
+    System.out.println(indent(driver.code))
+  }
```
feiwang3311, line 426: The code generation of mutually recursive functions in Scala still keeps the lambda-forward symbol. can we fix it?

PR #19, opened 2020-06-16
[Extension] handle c library

feiwang3311:
initial effort for encapsulating c libraries in LMS. Thanks to @GSAir @Kraks for the guidances and comments. 
There are a few outstanding problems (as listed in the comments) but I have them summarized here:
0. Added a mustInline option in CompactTraverser. If the node op starts with 'INLINE', the node is enforced to be inlined. This is used for simple macros that are like constant.
1. maybe we need a stop-the-world effect to make sure `mpi-init` is at the very beginning of the  generated code (similar for `mpi-finalize`)
2. printf of char-array. We cannot directly allow `printf("%s", arr)` since it doesn't catch the read effect on `arr`. It makes sense to enforce the user code to be `printf("%s", arr.toString)`. 
2.1 How do we enforce a type error for code `printf("%s", arr)`
2.2 How to code-gen `arr.toString` (with a copy or just slice?, is it target-language specific?)


src/main/scala/lms/core/backend.scala
```diff
@@ -522,6 +522,11 @@ class GraphBuilder {
 
 case class Graph(val nodes: Seq[Node], val block: Block, val globalDefsCache: immutable.Map[Sym,Node]) {
   // contract: nodes is sorted topologically
+  def show() = {
```
Kraks, line 525:
```suggestion
  def show: Unit = {
```

src/main/scala/lms/thirdparty/mpi.scala
```diff
@@ -0,0 +1,76 @@
...
+
+import lms.core._
+import lms.util._
+import lms.core.stub._
+import lms.core.Backend._
+import lms.core.virtualize
+import lms.core.utils.time
+import lms.macros.SourceContext
+
+trait MPIOps extends lms.collection.PointerOps { b: Base =>
```
Kraks, line 11:
```suggestion
import lms.collection._
trait MPIOps { b: Base with PointerOps =>
```

src/main/scala/lms/collection/PointerOps.scala
```diff
@@ -0,0 +1,42 @@
...
+import scala.collection.mutable.HashMap
+
+trait PointerOps { b: Base =>
+  object Pointer {
+    def apply[A: Manifest](x: Var[A])(implicit pos: SourceContext) = {
+      val a = Wrap[Pointer[A]](Adapter.g.reflect("pointer-new", UnwrapV(x)))
+      cache += ((Unwrap(a), UnwrapV(x)))
+      a
+    }
+    def deref[A: Manifest](x: Rep[Pointer[A]])(implicit pos: SourceContext) = {
```
Kraks, line 19: I think `deref` is better to be a method of something like `PointerOps` (which provides the syntax and operations you can do on a pointer instance), instead of in the singleton object `Pointer`.

src/main/scala/lms/core/codegen/C_codegen.scala
```diff
@@ -396,6 +396,8 @@ class ExtendedCCodeGen extends CompactScalaCodeGen with ExtendedCodeGen {
     case n @ Node(s,"array_resize", List(arr, nsize), _) =>
       val tpe = remap(typeMap.get(s).map(_.typeArguments.head).getOrElse(manifest[Unknown]))
       emit(s"($tpe*)realloc("); shallow(arr); emit(", "); shallow1(nsize, precedence("*")); emit(s" * sizeof($tpe))")
+    case n @ Node(s, "array-of-char-to-string", List(x:Sym), _) =>
+      shallow(x)
```
feiwang3311, line 400: this is currently wrong. Should have an array-copy first, the return the copied array. However, array_length node code-gen is still missing, making array-copy unfeasible. Will fix later.

src/main/scala/lms/core/stub.scala
```diff
@@ -761,6 +761,12 @@ trait Base extends EmbeddedControls with OverloadHack with lms.util.ClosureCompa
     def sort(size: Rep[Int] /* , sort: (Rep[A], Rep[A]) => Rep[Int] */) = Wrap[Array[A]](Adapter.g.reflectMutable("array_sort_scala", Unwrap(x), Unwrap(size))) //, Adapter.g.reify((x, y) => Unwrap(sort(Wrap[A](x), Wrap[A](y)))))(Unwrap(x))(Unwrap(x))
   }
 
+  implicit class CharArrayOps(x: Rep[Array[Char]]) {
+    def ArrayOfCharToString(): Rep[String] = {
+      Wrap[String](Adapter.g.reflectRead("array-of-char-to-string", Unwrap(x))(Unwrap(x)))
```
feiwang3311, line 766: should do array-copy and then array-to-string

src/main/scala/lms/core/stub.scala
```diff
@@ -926,7 +932,7 @@ trait Base extends EmbeddedControls with OverloadHack with lms.util.ClosureCompa
     Adapter.g.reflectWrite("P",Unwrap(x))(Adapter.CTRL)
 
   def printf(f: String, x: Rep[Any]*): Unit = {
-    Adapter.g.reflectWrite("printf",Backend.Const(f)::x.map(Unwrap).toList:_*)(Adapter.CTRL)
+    Adapter.g.reflectEffect("printf", Backend.Const(f)::x.map(Unwrap).toList: _*)()(Adapter.CTRL)
```
feiwang3311, line 935: this change is useless, but I am pinning it here because I want to say that we need to enforce a type error if users try to printf an array of char

src/main/scala/lms/collection/PointerOps.scala
feiwang3311: make sense. However, it is actually designed to not be used by users (only used to get the accurate read/write effects to the correct Sym). Does this make sense? Maybe it should be made private somehow?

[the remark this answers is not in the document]
Kraks: Well, then perhaps it needs a better name. But anyway, with an implicit `PointerOps` class, it would be more natural to write `p.deref` instead `Pointer.deref(p)`.

src/main/scala/lms/collection/PointerOps.scala
```diff
@@ -0,0 +1,41 @@
...
+trait PointerOps { b: Base =>
+  object Pointer {
+    def apply[A: Manifest](x: Var[A])(implicit pos: SourceContext) = {
+      val a = Wrap[Pointer[A]](Adapter.g.reflect("pointer-new", UnwrapV(x)))
+      cache += ((Unwrap(a), UnwrapV(x)))
+      a
+    }
+    def deref[A: Manifest](x: Rep[Pointer[A]]) = {
+      Wrap[A](cache.getOrElse(Unwrap(x), ???))
+    }
```
feiwang3311, line 21: this is not real `deref`of pointers, but just a way to back-track the Sym for read/write effects. Would a different name be more appropriate.

src/main/scala/lms/core/stub.scala
feiwang3311: f.split("%").map(_,charAt(0)) zip(x.map(typemap)) foreach {
  case ('d', m) => assert(m == manifest[Int])
  ...
}

or

-- Parse the format string (list of characters) into an AST.
-- Example: "%d,%s"  →   (FInt (FOther ',' (FString FEnd)))
format : List Char -> Format
format ('%' :: 'd' :: cs ) = FInt ( format cs )
format ('%' :: 's' :: cs ) = FString ( format cs )
format ( c :: cs )         = FOther c ( format cs )
format []                  = FEnd

https://gist.github.com/chrisdone/672efcd784528b7d0b7e17ad9c115292

src/main/scala/lms/collection/PointerOps.scala
```diff
@@ -0,0 +1,48 @@
...
+      cache += ((Unwrap(a), Unwrap(x)))
+      a
+    }
+    def deref[A: Manifest](x: Rep[Pointer[A]]) = {
+      Wrap[A](cache.getOrElse(Unwrap(x), ???))
+    }
+    private[this] val cache = HashMap[lms.core.Backend.Exp, lms.core.Backend.Exp]()
+  }
+
+  abstract class Pointer[T] extends Manifest[T]
```
Kraks, line 30: what does it mean by extending Manifest[T]?

feiwang3311: I am not very sure 😂 . 

I thought since we do `Wrap[+A:Manifest]`, then all Rep[A] must have A <: Manifest, so I thought Pointer might also needs to extends Manifest. But just tried and found no compilation error, so looks like it is not necessary.

Kraks: ok, but there needs a Manifest for T, right? like
`abstract class Pointer[T : Manifest]`

feiwang3311: Em. I agree with you. Strangely, it has not given me any compilation errors. Will fix.

src/main/scala/lms/thirdparty/mpi.scala
```diff
@@ -0,0 +1,76 @@
...
+  /* LMS support for MPI library */
+
+  def mpi_init(): Rep[Unit] = Wrap[Unit](Adapter.g.reflectWrite("mpi-init")(Adapter.CTRL))
```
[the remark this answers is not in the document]
GSAir: Is it something you witnessed or you are afraid that would happen? The code isn't moving around freely, so we know when it will move. What mechanism would make that statement move? If I write code that does:

```
def main() = {
   mpi_init()
   // ...
}
```

There is absolutely now reason that this statement won't be first. The code motion mechanisms are:
- inlining: not going to happen with that one.
- move non effectful statement in if block: not going to happen
- take non effectful statement out of a while/lambda block: not going to happen.

src/main/scala/lms/thirdparty/mpi.scala
```diff
@@ -0,0 +1,83 @@
...
+  override def mayInline(n: Node): Boolean = n match {
+    case Node(_, "datastructure1-new", _, _) => false
+    case _ => super.mayInline(n)
+  }
+
+  override def shallow(n: Node): Unit = n match {
+    case Node(s, "macro-mutate", List(x), _) =>
+      emit("MUTATE("); shallow(x); emit(")")
+    case Node(s, "INLINE_mpi-comm-world", _, _) =>
+      emit("MPI_COMM_WORLD")
```
GSAir, line 62: Why do we have a Node for a constant? That's equivalent to have a Node for 1, 2, 3, 4...

src/main/scala/lms/thirdparty/mpi.scala
```diff
@@ -0,0 +1,83 @@
...
+import lms.core._
+import lms.util._
+import lms.core.stub._
+import lms.core.Backend._
+import lms.core.virtualize
+import lms.core.utils.time
+import lms.macros.SourceContext
+
+import lms.collection._
+trait MPIOps { b: Base with PointerOps =>
```
GSAir, line 12: Any reason why you don't just extend the trait instead of using that?

src/main/scala/lms/thirdparty/mpi.scala
```diff
@@ -0,0 +1,83 @@
...
+  /* LMS support for MPI library */
+
+  def mpi_init(): Rep[Unit] = Wrap[Unit](Adapter.g.reflectWrite("mpi-init")(Adapter.CTRL))
+
+  abstract class MPIWorld extends Manifest[MPIWorld]
+  lazy val mpi_comm_world: Rep[MPIWorld] = Wrap[MPIWorld](Adapter.g.reflect("INLINE_mpi-comm-world"))
```
GSAir, line 18: Why are we using a reflect for a constant?

```
object MPICommWorld {
   override def toString = "MPI_COMM_WORLD"
}
val mpi_comm_world: Rep[MPIWorld] = unit(MPICommWorld)
```

I don't really like the _ in names but I guess that's ok. I would match the exact C name if we wanted to keep the convention, or completely making Scala like. This seems like a middle ground.

src/main/scala/lms/collection/PointerOps.scala
GSAir: I would look into the existing support for `Pointer` I have in [Flare](https://github.com/GSAir/Flare/blob/lms_clean/src/main/scala/flare/common/Record.scala#L143).

The code you have here shouldn't be a LMS trait. It is not correct at all. I am not sure what you are expecting to solve with the cache, but that's not going to work in the general case. What is the push to have this code in LMS? Is it only for MPI and the fact we have to pass pointers to function call? That can be handle at the Node level for MPI directive, I don't think we should add an incomplete code for something that is not really a problem in the first place.

So I went ahead and I see that:
```
def mpi_comm_size(world: Rep[MPIWorld], size: Rep[Pointer[Int]]): Rep[Unit] = {
    Wrap[Unit](Adapter.g.reflectWrite("mpi-comm-size", Unwrap(world), Unwrap(size))(Unwrap(Pointer.deref(size))))
}
```
Why would we want code that looks like that? This is just a mapping to C code.

```
def mpi_comm_size(world: Rep[MPIWorld]): Rep[Int] = {
    Wrap[Int](Adapter.g.reflect("mpi-comm-size", Unwrap(world))
}

// in traverse:
...
   case Node(s, "mpi-comm-size", List(arg), ....) =>
         emitln("${remap(manifest[Int])} ${quote(s)};") 
        emit("MPI_COMM_SIZE("); shallow(arg); emit(", & ${quote(s)}")")
```

it shouldn't be inlined, and there may be a small issue on formatting. The quoteBlock functions assume there is a single line generated per Node I think. Do you see the idea? I don't see any reason to have an effectful statement on a wonky pointer representation.

I think we should have a better support but the solution with the cache is not going to make it. The one in Flare isn't perfect either, but it is required for performance and I know that it is only use correctly within Flare. Adding it to LMS would mean that it can be used everywhere. And the simple counter example is:

```
var x1 = 0
var x2 = 0
val ptr = if (in == 0) Pointer(x1) else Pointer(x2)
mpi_comm_size(..., ptr)
```

Maybe there is a usage I don't see.

src/main/scala/lms/core/traversal.scala
```diff
@@ -263,6 +263,8 @@ abstract class CPSTraverser extends Traverser {
 
 class CompactTraverser extends Traverser {
 
+  val mustInline = mutable.HashMap[Sym, Node]()
```
GSAir, line 266: That's scary

src/main/scala/lms/thirdparty/mpi.scala
```diff
@@ -0,0 +1,83 @@
...
+  def mpi_get_processor_name(name: Rep[Array[Char]], len: Rep[Pointer[Int]]): Rep[Unit] = {
+    Wrap[Unit](Adapter.g.reflectWrite("mpi-get-processor-name", Unwrap(name), Unwrap(len))
+    (Unwrap(name), Unwrap(Pointer.deref(len))))
+  }
+
+  def mpi_finalize(): Rep[Unit] = Wrap[Unit](Adapter.g.reflectWrite("mpi-finalize")(Adapter.CTRL))
+
+  object DataStructure1 {
+    def apply()(implicit pos: SourceContext): Rep[DataStructure1] = {
+      Wrap[DataStructure1](Adapter.g.reflect("datastructure1-new"))
```
GSAir, line 39: I am scared of what that is. Look at the refined manifest in Flare. There is already (better) support for that.

src/main/scala/lms/collection/PointerOps.scala
feiwang3311: Thanks Greg. You are right that this is not correct pointer. The motivation is to not hard-code the `&` when the library function (not just MPI though) needs pointers. It can also be handled in the way you suggested `emit(", & ${quote(s)}")"`, with the `mpi_comm_size(world: Rep[MPIWorld]): Rep[Int]` function hiding the C library call details.

I guess my approach allows more flexibility. When a library function wants int* parameter, it could be taking `&x1` if `x1` is int, or `x2` if `x2` is int array. That means not all applications needs to use `&`. But maybe I went to far since generally the library functions are used in a fixed manner.

src/main/scala/lms/core/codegen/C_codegen.scala
feiwang3311: not sure I follow what the flare code is doing. :) Is it building a string?

I was trying to copy the array but hit a ??? when fetch array-length. I guess that needs to be fixed anyway.

src/main/scala/lms/core/stub.scala
feiwang3311: this code needs to be fixed with format string checks. :) 

not intending to push bad code in the repo :)

src/main/scala/lms/core/traversal.scala
feiwang3311: This is for library Macros that serves as constant (such as MPI_COMM_WORLD). It is not wrong if they are not inlined, but more correct if they are.

src/main/scala/lms/thirdparty/mpi.scala
feiwang3311: no reason :) Will do extends :)

src/main/scala/lms/thirdparty/mpi.scala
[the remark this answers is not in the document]
feiwang3311: Thanks Greg. Your analysis assures that the code will be at the top of the snippet. 

But do we need a `stop-of-the-world` effect anyway? If so, how do you suggest we do it?

src/main/scala/lms/thirdparty/mpi.scala
feiwang3311: this is super interesting. Seems a big simplification to using `reflect` and solves the `mustInline` problem above :) 

In order for the type to be correct, I had to make some modifications
  
```
  class MPIWorld extends Manifest[MPIWorld] {
    def runtimeClass = classOf[MPIWorld]
  }

  object MPICommWorld extends MPIWorld {
    override def toString = "MPI_COMM_WORLD"
  }
  lazy val mpi_comm_world: Rep[MPIWorld] = unit(MPICommWorld)
```
does this change make sense?

src/main/scala/lms/thirdparty/mpi.scala
feiwang3311: will remove :)

src/main/scala/lms/thirdparty/mpi.scala
feiwang3311: will remove :)

src/main/scala/lms/thirdparty/mpi.scala
feiwang3311: Can you elaborate? This is trying to support code gen like 
```
DataStructure1 datas;
...
```
not sure how refined manifest is helping

src/main/scala/lms/core/codegen/C_codegen.scala
GSAir: No... this is not defined in C, so there is no implementation possible. I am just showing you functions that have the length given as argument.

src/main/scala/lms/thirdparty/mpi.scala
```diff
@@ -0,0 +1,87 @@
...
+  }
+  def test_struct_reference(x: Rep[DataStructure1]): Rep[Unit] = {
+    Wrap[Unit](Adapter.g.reflectWrite("test-struct-ref", Unwrap(x))(Adapter.CTRL, Unwrap(x)))
+  }
+}
+
+trait CCodeGenMPI extends ExtendedCCodeGen {
+  override def remap(m: Manifest[_]): String =
+    m.runtimeClass.getName match {
+      case "lms.thirdparty.MPIOps$DataStructure1" => "DataStructure1"
```
feiwang3311, line 60: this case of `remap` is still needed, but it is simply because one of the test defines a new `topFun` using the type of the `DataStruct1`. This rarely happens in library bindings :)

PR #22, opened 2020-06-22
[CodeGen] cpp codegen with closure

feiwang3311:
1. add CPPCodeGen
2. move and enrich closure code gen from CCodeGen to CPPCodeGen
3. tests and checked files
4. more re-organization of stub.scala

Outstanding Problem:
In handling closure codeGen of CPP, we replace the lambda_forward symbol with lambda. 
that is to say: if x2 = lambda_forward(x3), then replace all codegen of x2 with x3. 
This is current wrong because it generate invalid code for mutually recursive closures.


src/main/scala/lms/core/backend.scala
```diff
@@ -511,6 +511,190 @@ class GraphBuilder {
   }
 }
 
+class GraphBuilderOpt extends GraphBuilder {
```
feiwang3311, line 514: moved this class from Adapter object to backend.scala

src/main/scala/lms/core/codegen/C_codegen.scala
```diff
@@ -148,7 +148,7 @@ class ExtendedCCodeGen extends CompactCodeGen with ExtendedCodeGen {
     }
     withWraper(wraper _)(f)
   }
-  def quoteBlock(b: Block): Unit = quoteTypedBlock(b, true, false)
+  def quoteBlock(b: Block): Unit = ???
```
feiwang3311, line 151: in c code gen, it should not be allowed to generate closure. 
this is moved to cpp code gen.

src/main/scala/lms/core/codegen/C_codegen.scala
```diff
@@ -210,31 +210,6 @@ class ExtendedCCodeGen extends CompactCodeGen with ExtendedCodeGen {
     withWraper(wraper _)(f)
   }
 
-  // emit closures with explicit argument types and return type
```
feiwang3311, line 213: moved to cpp code gen

src/main/scala/lms/core/codegen/cpp_codegen.scala
```diff
@@ -0,0 +1,102 @@
...
+  }
+  case class Share(s: String) extends Capture
+  case class UniqueCopy[T:Manifest](s: Rep[T]) extends Capture
+  case class UniqueRef[T:Manifest](s: Rep[T]) extends Capture
+  implicit def liftCapture(s:String) = Capture(s)
+  private def unwrap(c: Capture) = c match {
+    case Share(s: String) => lms.core.Backend.Const(s)
+    case UniqueCopy(s: Rep[_]) => Unwrap(s)
+    case UniqueRef(s: Rep[_]) => Unwrap(s)
+  }
```
feiwang3311, line 28: allowing programmer to explicitly control `capture` behavior (via copy or via reference)

src/main/scala/lms/core/stub.scala
```diff
@@ -75,203 +69,18 @@ object Adapter extends FrontEnd {
       cg.extractAllStatics
     }
   }
-
-  override def mkGraphBuilder() = new MyGraphBuilder()
```
feiwang3311, line 79: moved to backend.scala

src/main/scala/lms/core/stub.scala
```diff
@@ -656,128 +465,6 @@ trait Base extends EmbeddedControls with OverloadHack with lms.util.ClosureCompa
   // def implicit_convert[A:Manifest,B:Manifest](a: Rep[A]): Rep[B] = Wrap[B](Adapter.g.reflect("convert", Unwrap(a)))
 }
 
-
-trait Dsl extends PrimitiveOps with LiftPrimitives with Equal with RangeOps with OrderingOps with lms.collection.mutable.ArrayOps with UtilOps {
```
feiwang3311, line 660: moved to the end of this file

src/main/scala/lms/core/stub.scala
```diff
@@ -1367,6 +1052,108 @@ trait PrimitiveOps extends Base with OverloadHack {
...
+      else s"-I ${codegen.includePaths.mkString(" -I ")}"
+    val pb: ProcessBuilder = s"$compilerCommand /tmp/snippet.c -o /tmp/snippet $libraries $includes"
+    time("gcc") { pb.lines.foreach(Console.println _) }
+    (a: A) => (s"/tmp/snippet $a": ProcessBuilder).lines.foreach(Console.println _)
+  }
+  def eval(a: A): Unit = { val f1 = f; time("eval")(f1(a)) }
+}
+
+// Basic DslDriverCPP for CPP CodeGen
+abstract class DslDriverCPP[A: Manifest, B: Manifest] extends DslDriverC[A, B] with CPPOps { q =>
```
feiwang3311, line 1150: DslDriverCPP is an extension of DslDriverC

src/main/scala/lms/core/codegen/cpp_codegen.scala
```diff
@@ -0,0 +1,102 @@
...
+
+  abstract class Capture
+  object Capture {
+    def apply(s: String) = Seq(Share(s))
+    def CopyExcept[T:Manifest](s: Rep[T]*) = Seq(Share("=")) ++ s.map(UniqueRef(_))
+    def RefExcept[T:Manifest](s: Rep[T]*) = Seq(Share("&")) ++ s.map(UniqueCopy(_))
+  }
+  case class Share(s: String) extends Capture
+  case class UniqueCopy[T:Manifest](s: Rep[T]) extends Capture
+  case class UniqueRef[T:Manifest](s: Rep[T]) extends Capture
```
Kraks, line 22: I think I can guess the meaning of `Share`, `UniqueCopy` and `UniqueRef`, but It seems not an intuitive mapping to conventions (by-copy and by-ref) unless they have other meanings. Especially, what does "unique" mean here? they remind me shared_ptr and unique_ptr which is confusing.

src/main/scala/lms/core/codegen/cpp_codegen.scala
```diff
@@ -0,0 +1,102 @@
...
+trait CPPOps { b: Base =>
+
+  abstract class Capture
+  object Capture {
+    def apply(s: String) = Seq(Share(s))
+    def CopyExcept[T:Manifest](s: Rep[T]*) = Seq(Share("=")) ++ s.map(UniqueRef(_))
+    def RefExcept[T:Manifest](s: Rep[T]*) = Seq(Share("&")) ++ s.map(UniqueCopy(_))
+  }
+  case class Share(s: String) extends Capture
+  case class UniqueCopy[T:Manifest](s: Rep[T]) extends Capture
```
Kraks, line 21:
```suggestion
  case class UniqueCopy[T: Manifest](s: Rep[T]) extends Capture
```

src/main/scala/lms/core/codegen/cpp_codegen.scala
Kraks, line 22:
```suggestion
  case class UniqueRef[T: Manifest](s: Rep[T]) extends Capture
```

src/main/scala/lms/core/codegen/cpp_codegen.scala
```diff
@@ -0,0 +1,102 @@
...
+import Backend._
+import lms.core.stub._
+
+trait CPPOps { b: Base =>
+
+  abstract class Capture
+  object Capture {
+    def apply(s: String) = Seq(Share(s))
+    def CopyExcept[T:Manifest](s: Rep[T]*) = Seq(Share("=")) ++ s.map(UniqueRef(_))
+    def RefExcept[T:Manifest](s: Rep[T]*) = Seq(Share("&")) ++ s.map(UniqueCopy(_))
```
Kraks, line 18:
```suggestion
    def CopyExcept[T: Manifest](s: Rep[T]*) = Seq(Share("=")) ++ s.map(UniqueRef(_))
    def RefExcept[T: Manifest](s: Rep[T]*) = Seq(Share("&")) ++ s.map(UniqueCopy(_))
```

src/main/scala/lms/core/codegen/cpp_codegen.scala
feiwang3311: the user is not supposed to use them, but rather use the functions in `object Capture`.

I tried to make the case classes `private` but it doesn't compile :)

src/main/scala/lms/core/codegen/C_codegen.scala
TiarkRompf: It depends ... gcc has support for nested functions. There should probably be some kind of flag to enable or disable gcc/clang specific patterns. We already use some other patterns like statement expressions that aren't part of ANSI C. It's nice to have them available because the generated code is more compact (and often more readable) but we have to be able to turn these off for less-fancy compilers (importantly, TCC).

In plain C we can add a closure conversion phase (I believe @GSAir implemented parts of it for OSR)

Not saying all this needs to be done right now, perhaps let's create a ticket to remind us later.

feiwang3311: Interesting! I didn't know that gcc support nested functions :) 

But so far, `quoteTypedBlock` goes to closure generation (via syntax: `[&](int a){ ... }), which needs to be fixed I think (we should use normal function syntax such as `int f(int x) {..}`

I also tested and found that the nested function can access variables defined out of the scope, such as ```
int main()
{
    int x = 5;
    int inner(int a) { 
        return a + x; 
    }
    printf("%d\n", inner(3));
    return 0;
}```
So this is not so different from closures IMO. Are there any difference between gcc-supported nested functions and c++ closures? 

It seems that there are a few things to handle here:
1. flag to support inner functions or not (decided based on if we use gcc or tcc)
2. when supporting inner functions, generate them in function syntax `int f(int x) { ... }`
3. when not supporting inner functions, closure conversions
4. for cpp codegen compiled with gcc, it seems that we can either use cpp closure or inner functions (not sure if there are differences between them). How do we choose?
@TiarkRompf @GSAir @Kraks

src/main/scala/lms/core/codegen/cpp_codegen.scala
GSAir: I think he meant for us the programmer. The fact that users don't have access doesn't mean we shouldn't give them name that makes sense.

src/main/scala/lms/core/codegen/cpp_codegen.scala
GSAir: Capital letter for functions? ++?

src/main/scala/lms/core/codegen/cpp_codegen.scala
feiwang3311: fixed :)

src/main/scala/lms/core/codegen/C_codegen.scala
Kraks: When targeting modern C++, I think there is no need to generate nested functions (in fact it is not supported by GNU C++ according to its document).

PR #24, opened 2020-06-22
[Extension] better type handling

feiwang3311:


Kraks:
Not sure if it would be helpful, but I'd suggest that maybe we can borrow some API design from Scala Native.

src/main/scala/lms/thirdparty/mpi.scala
```diff
@@ -37,13 +37,21 @@ trait MPIOps extends CMacro with LibStruct with LibFunction { b: Base =>
   }
 
   // this is how we bind to library structs with field read access
+  // 1. define the abstract class (if not abstract, works too)
   abstract class DataStructure1
+  // 2. handle type mapping by CustomManifest[T] with the implicit object and toString method
+  implicit object ManifestDataStructure1 extends CustomManifest[DataStructure1] {
```
Kraks, line 43: It seems the codegen `remap` will use whatever in the `toString` as the type name for this struct. But, to me, I think it couples the backend-sepcific struct name with front-end syntax/IR too early. Imagine that at some later point, the DSL developer wants to change the backend, then the developer has to not only update the codegen but also the front-end `XOps`. Ideally, the front-end syntax/operations and IR should be backend irrelevant, and `Manifest[T]` is the descriptor of a _Scala_ type `T`, not the descriptor of its corresponding _C_ type.

feiwang3311: Interesting counter-argument. 

I guess for this one, we are binding to a specific library with known syntax. So the concern of separating frontend and codegen for potentially multiple backend is not so relevant? I mean, the MPI API should have a fix syntax for various hardware-systems, right?

Kraks: > the MPI API should have a fix syntax for various hardware-systems, right?

Right, for a specific case, there are many ways to do it. But what we are looking for is simple abstraction and principles that apply to all cases.

I can see the proposed principle behind this PR is to use CustomManifest to represent the descriptor of both the Scala type and the corresponding next-stage type, so that the codegen can use the type name whatever it is. However, if there are multiple backends, should we create multiple `CustomManifest` for different backends? Or should we further add multiple backend type strings in a single `CustomManifest`? In my opinion, they all complicate things.

Because the manifest information is needed in frontend operations, adding the CustomManifest and adding the next-stage type string breaks the abstraction between frontend operations and codegen, also makes things less _modular_.

The existing approach is to use `remap` to transform the descriptor of a Scala type `T` to next-stage type (represented by string) during codegen (it happens very late). I think this should work for this MPI case, if CustomManifest doesn't have any other purpose. What I felt less comfortable is there are two contradictory ways to do the same thing in a single framework.

> Sometimes people forget to update remap function for a struct.

Then they may also forget to update or provide CustomManifest, nothing prevent that :)

feiwang3311: I like your argument :) @GSAir What do you think?

src/main/scala/lms/macros/CustomManifest.scala
```diff
@@ -0,0 +1,5 @@
+package lms.macros
+
+trait CustomManifest[T] extends Manifest[T] {
```
GSAir, line 3: That's not what I meant. And also why the runtime class would be an Int?
And why is that in macro?

src/main/scala/lms/thirdparty/mpi.scala
GSAir: I think that this shouldn't be done that way. I would have a function, like regsterType, that does all the mechanic rather than letting the user having to do it.

But I disagree that it links back and frontend. The Manifest is still used to represent the type, that doesn't change. We just use the String representation for the backend. Overriding remap is exactly the same thing. I don't see how having a CMacro frontend can be modular, this is only going to work with C whatever we do.

src/main/scala/lms/macros/CustomManifest.scala
feiwang3311: fixed :)

src/main/scala/lms/thirdparty/mpi.scala
```diff
@@ -37,13 +37,18 @@ trait MPIOps extends CMacro with LibStruct with LibFunction { b: Base =>
...
   // this is how we bind to library structs with field read access
-  abstract class DataStructure1
+  class DataStructure1
+  // 1. define an implicit value for c-type registration (internally it is a CustomManifest)
+  implicit val registCTypeForDataStructure1 = registerCType[DataStructure1]("DataStructure1")
+  // 2. define the instantiation function
+  def dataStructure1: Rep[DataStructure1] = newStruct[DataStructure1]
+  // 3. add any field read access functions
   implicit class DataStructure1Ops(x: Rep[DataStructure1]) {
     val fieldA: Rep[Int] = readField[DataStructure1, Int](x, "fieldA")
```
Kraks, line 47: is it should be a `def`?

feiwang3311: better to be a `def` I think :) 

using `val` is assuming that no more changes happen to the struct, which is sort of true here (and helps with CSE), but quite constraint.

feiwang3311: fixed

PR #25, opened 2020-06-23
[Backend] fun: partially fix lambda forward effect

feiwang3311:
Previously `lambda forward` just has `CTRL` latent effect, which might intend to be `stop-the-world-effect` but is not really. I wrote a test (added in this PR) to probe the wrong behavior.

I am not sure what the right way to fix it is:
1. accurate computation of lambda forward latent effect seems overly complex
2. a real stop-of-the-world effect is needed, but how to implement it?
3. temp solution in this PR is to enforce read/write effects to all arguments. I am composing it as a partial step toward a `stop-of-the-world` effect, and also proves that it fixes my test (and probably some Lantern tests as well).
4. should be able to be precise or conservative or optimistic (be able to choose).

### DIFF
--- a/src/main/scala/lms/core/backend.scala
+++ b/src/main/scala/lms/core/backend.scala
@@ -369,8 +369,8 @@ class GraphBuilder {
   def getFunctionLatentEffect(f: Exp): ((Set[Exp], Set[Exp]),(Set[Int], Set[Int]), Option[Exp]) = findDefinition(f) match {
       case Some(Node(_, "λ", (b:Block)::_, _)) =>
         getEffKeysWithParam(b)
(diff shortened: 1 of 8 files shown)


Kraks:
what exactly is `stop-the-world` effect?

feiwang3311:
> what exactly is `stop-the-world` effect?

we have various categories of effect keys: one category for each var or array, STORE, CTRL, UNSAFE, et al.

At some point, we have so little information about the node, such that we want the node to have read and write effects to all categories of effect keys. The expected behavior is that we don't move any code across the node with `stop-of-the-world` effect, and no dead-code-elimination happens. With `stop-of-the-world` effect, we are making conservative decisions to generate correct code.

PR #28, opened 2020-06-28
[CudaCodeGen] support cuda function application syntax

feiwang3311:


src/main/scala/lms/thirdparty/c_lib_utils.scala
```diff
@@ -52,6 +52,12 @@ trait CCodeGenLibStruct extends ExtendedCCodeGen {
       emit(s"${remap(m)} ")
       shallow(s)
       emitln(";")
+    // allow "lib-struct" to take parameters
+    case Node(s, "lib-struct", Const(m: Manifest[_])::args, _) =>
+      emit(s"${remap(m)} "); shallow(s);
+      emit("("); shallow(args.head);
+      args.tail.foreach(a => {emit(", "); shallow(a)})
```
Kraks, line 59: Does this output code like this?
```
struct Point x(arg1, arg2, ...);
```

feiwang3311: yes :)

Kraks: is that a valid way to initialize a struct in C?

feiwang3311: Sorry, it should generate `Point x(arg1, arg2, ...)`.

There is no `struct` at the begining.

Kraks: Hmm, if `Point` is typedef-defined, eg:
```
typedef struct Point { int x; int y; } Point;
```
that seems doesn't change anything. IIRC, writing `Point p(1, 2);` seems still not valid C...

feiwang3311: Right. It works in C++ if the struct also defines a constructor of that type. 

I guess we don't need to worry about the `defines a constructor` part, since the users should know that from the library.

Do you think it makes more sense to put this in CPP_lib_utils.scala?

Kraks: If the generated code is C++, it indeed should be in C++ code gen. But before doing that, I think we might need to think more about supporting `struct`/`class` in general...

- I guess we still need to generate C code that initialize `struct` here, like:
```
<struct_type> <var> = { .<field1> = <val1>, ... };
``` 
or simply
```
<struct_type> <var> = { <val1>, ... };
```

- The current `lib-struct` essentially expresses opaque `struct`, that is we only know the type name of this `struct` but not the field names and structure. Most of the cases, the DSL user knows (or can know) the structure and field names. The opaque `struct` is useful when it is not necessary to manipulate the fields at meta level. I wonder if this can be unified with the existing `Record`.

PR #32, opened 2020-07-17
[LMS-clean] be careful with using lazy val

feiwang3311:


src/test/scala/lms/core/test_lambda.scala
```diff
@@ -437,4 +437,23 @@ class LambdaTest extends TutorialFunSuite {
...
     check("mutual_recursion", driver.code, "scala")
   }
+
+  test("be_careful_with_lazy") {
+    val driver = new DslDriver[Int, Unit] {
+      @virtualize
+      def snippet(a: Rep[Int]) = {
+        // if the fun1 is `lazy val` the generated code has
+        // 2 instances of the fun2 code, since the `lazy val`
+        // semantics changes the canonicanize result of `fun2`
```
GSAir, line 447: That's not the correct explanation. `lazy val` can't change the canonize result. `lazy` doesn't change the representation of the object...

```
lazy val fun1 = ...

val fun2 = fun { (x: Rep[Int]) =>
   fun1(x/2)
}
```

is "equivalent to"

```
var fun1 = null
val fun2 = fun { (x: Rep[Int]) =>
   fun1 = ...
   fun1(x/2)
}
```

for the first call to `fun2` and 

```
val fun2 = fun { (x: Rep[Int]) =>
   fun1(x/2)
}
```
for the second. I think that's why it is changing the canonize result. It is because `fun2` is a `def` (why????)

feiwang3311: I have printed the canonization result to show that they are different.

basically if obj A has B inside, and B is a lazy val, then the canonizations of A `before B is evaluated` and `after B is evaluated` are different. 

the fun2 being def is OK. For all recursive lambda in LMS, we should be able to use `lazy val` and `def` interchangeably without affecting the code generation.

GSAir: Yes that makes sense that they are different. But because it is a `def`. It is producing the function multiple times.

What is the goalof that test? What is it testing? What does "be careful with lazy" means?

Or this is a programming advice to avoid problems, and that's not a test. Or we change the code to make this code works, but I am not sure what's the test is supposed to test here.

feiwang3311: yeah, the test is just saying that we cannot use `lazy val` on `fun1`. 

people can try to change it to `lazy val` and see how the code gen is different :) 

for `fun2`, we cannot use `val` because it is recursive. We can use either `def` or `lazy val`, and that will be no difference, since the `__fun` function has a caching mechanism (via canonicalization and funTable).

PR #35, opened 2020-07-19
[CodeGen] flexible target name

feiwang3311:


src/main/scala/lms/thirdparty/c_lib_utils.scala
```diff
@@ -117,3 +118,63 @@ trait CudaCodeGenLibFunction extends ExtendedCCodeGen {
...
+}
+
+trait CCodeGenScannerOps extends ExtendedCCodeGen with CCodeGenLibs {
+
+  // need to register the headers
+  val workPath = System.getProperty("user.dir")
+  val headerPath = "src/main/scala/lms/thirdparty/thirdpartyAdaptor/"
+  val headerFile = "<scanner_header.h>"
+  registerHeader(s"$workPath/$headerPath", headerFile)
+  registerDefine("_GNU_SOURCE")
```
feiwang3311, line 172: somehow this `#define _GNU_SOURCE` is needed if we use some Macros related to file reading in C

src/main/scala/lms/thirdparty/c_lib_utils.scala
```diff
@@ -117,3 +118,66 @@ trait CudaCodeGenLibFunction extends ExtendedCCodeGen {
...
+  // mmap function maps a file to memory
+  def mmap[T:Manifest](fd: Rep[Int], len: Rep[Long]) = static_cast[Array[T], Array[T]](libFunction[Array[T]]("mmap",
+    lms.core.Backend.Const(0), Unwrap(len), Unwrap(prot), Unwrap(fd), lms.core.Backend.Const(0))(Seq[Int](), Seq[Int](), Set[Int]()))
+  def close(fd: Rep[Int]) = libFunction[Unit]("close", Unwrap(fd))(Seq[Int](), Seq[Int](), Set[Int](), Adapter.CTRL)
+  def prints(s: Rep[String]): Rep[Int] = libFunction[Int]("printll", Unwrap(s))(Seq[Int](), Seq[Int](), Set[Int](), Adapter.CTRL)
+
+  // establish a File* type
+  abstract class FilePointer
+  // another API for opening a file
+  def openf(name: Rep[String], mode: Rep[String]) = libFunction[FilePointer]("fopen", Unwrap(name), Unwrap(mode))(Seq[Int](), Seq[Int](), Set[Int](), Adapter.CTRL)
```
Kraks, line 145:
```suggestion
  def fopen(name: Rep[String], mode: Rep[String]) = libFunction[FilePointer]("fopen", Unwrap(name), Unwrap(mode))(Seq[Int](), Seq[Int](), Set[Int](), Adapter.CTRL)
```

src/main/scala/lms/thirdparty/c_lib_utils.scala
```diff
@@ -117,3 +118,66 @@ trait CudaCodeGenLibFunction extends ExtendedCCodeGen {
...
+  def closef(fp: Rep[FilePointer]) = libFunction[Unit]("fclose", Unwrap(fp))(Seq[Int](0), Seq[Int](), Set[Int](), Adapter.CTRL)
```
Kraks, line 146:
```suggestion
  def fclose(fp: Rep[FilePointer]) = libFunction[Unit]("fclose", Unwrap(fp))(Seq[Int](0), Seq[Int](), Set[Int](), Adapter.CTRL)
```

src/out/thirdparty/scanneropen_file.check.c
```diff
@@ -0,0 +1,28 @@
...
+#define _GNU_SOURCE
+#endif
+#include <stdlib.h>
+#include <scanner_header.h>
+#include <stdio.h>
+#include <stdint.h>
+#include <stdbool.h>
+/**************** Snippet ****************/
+void Snippet(int x0) {
+  int x1 = open("/scratch-ml00/wang603/lms-umbrella/lms-clean/src/test/scala/lms/thirdparty/test_scanner.scala", 0);
```
feiwang3311, line 14: this hard-coded path is wrong 🎃 

idk how to make this not hard-coded. 
maybe just don't check with files?

build.sbt
```diff
@@ -30,6 +30,7 @@ libraryDependencies += "org.scala-lang.plugins" % "scala-continuations-library_2
 
 scalacOptions += "-P:continuations:enable"
 
+val lms_path = sys.props.getOrElseUpdate("LMS_PATH", System.getProperty("user.dir"))
```
feiwang3311, line 33: use this to control the path to the hard-coded header file

src/main/scala/lms/core/backend.scala
```diff
@@ -587,8 +587,10 @@ class GraphBuilderOpt extends GraphBuilder {
     case ("array_slice", List(as: Exp, Const(0), Const(-1))) => Some(as)
     case ("array_length", List(Def("NewArray", Const(n)::_))) =>
       Some(Const(n))
-    case ("array_length", List(Def("Array", List(Const(as: Array[_]))))) =>
+    case ("array_length", List(Def("Array", as))) =>
       Some(Const(as.length))
+    case ("array_length", List(Def("array_slice", List(x, s, e)))) =>
+      Some(reflect("-", e, s))
```
feiwang3311, line 593: fix array_length for array_slice (used to be hitting ???)

Kraks: The change of `array_length`'s first case seems changing the behavior. If the creation of `array_length` remains the same, now `as.length` is always 1, is that right？

feiwang3311:
```  def Array[T:Manifest](xs: Rep[T]*): Rep[Array[T]] = {
    Wrap[Array[T]](Adapter.g.reflectMutable("Array", xs.map(Unwrap(_)):_*))
  }```

based on this code, we never store a list(const(array)), but a list of elements. 
the original code will never be triggered (unless it is array of array with just one element, and the result would be wrong 

Does this make sense?

Kraks: I see, it makes sense!

PR #38, opened 2020-07-26
[LMSclean] use resource to manage header

feiwang3311:


build.sbt
```diff
@@ -30,8 +30,6 @@ libraryDependencies += "org.scala-lang.plugins" % "scala-continuations-library_2
 
 scalacOptions += "-P:continuations:enable"
 
-val lms_path = sys.props.getOrElseUpdate("LMS_PATH", System.getProperty("user.dir"))
-
```
feiwang3311, line 34: remove old way to handle header files

src/main/scala/lms/core/codegen/C_codegen.scala
```diff
@@ -225,7 +225,32 @@ class ExtendedCCodeGen extends CompactCodeGen with ExtendedCodeGen {
...
+    if (f == "<cuda_header.h>") {
+      // this is a resouce in LMS clean
+      val curPath = System.getProperty("user.dir")
+      val path = new File(curPath + "/cuda_header.h")
+      val src = scala.io.Source.fromInputStream(getClass.getResourceAsStream("/cuda_header.h"))
+      val temp_out = new PrintWriter(path)
+      temp_out.println(src.getLines.mkString("\n"))
+      temp_out.flush
+      temp_out.close
+      registerIncludePath(curPath)
```
feiwang3311, line 250: this is where resource is generate.

There might be a question "why not generate the header files in /tmp?"
The answer is, in mlxx.cs.purdue.edu machine, we cannot use /tmp anymore :(

src/main/scala/lms/thirdparty/c_lib_utils.scala
```diff
@@ -89,92 +90,3 @@ trait CCodeGenLibFunction extends ExtendedCCodeGen {
...
-  val headerFile = "<scanner_header.h>"
-  registerHeader(s"$workPath/$headerPath", headerFile)
-  registerDefine("_GNU_SOURCE")
-
-  // type remap
-  override def remap(m: Manifest[_]) = m.runtimeClass.getName match {
-    case s: String if s.endsWith("FilePointer") => "FILE*"
-    case _ => super.remap(m)
-  }
-}
```
feiwang3311, line 180: moved to separate files

src/test/scala/lms/thirdparty/test_size_t.scala
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  test("basic") {
+    val driver = new DslDriverCSizeT[Int, Unit] {
+      @virtualize
+      def snippet(arg: Rep[Int]) = {
+        val f = topFun { a: Rep[SizeT] =>
+          printf("size t is %d", a.toInt)
+        }
+        f(SizeT(5))
```
Kraks, line 25: Does `f(5)` work here?

src/main/scala/lms/collection/ArrayOps.scala
```diff
@@ -64,3 +64,35 @@ trait ArrayOps { b: Base =>
 }
 
 // FIXME(feiw) CodeGen of Array is still not yet separated from ExtendedScalaCodeGen and ExtendedCCodeGen
+
+// StackArray is the array using stack (i.e. int a[5] = {1,2,3,4,5};)
+trait StackArrayOps extends ArrayOps { b: Base =>
+
+  def NewStackArray[T:Manifest](x: Rep[Int]): Rep[Array[T]] = {
```
Kraks, line 71: Can `NewStackArray` and `StackArray` use the same name?

feiwang3311: probably not. The parameter in NewStackArray is for size. The parameter for StackArray is for content. There might be ambiguity if they have the same name?

src/test/scala/lms/thirdparty/test_size_t.scala
feiwang3311: it is not working yet because you would need 2 level of implicit conversion to get Rep[SizeT] from Int

PR #39, opened 2020-07-28
[ThirdParty] cuda and cublas

feiwang3311:


src/main/scala/lms/core/stub.scala
```diff
@@ -577,26 +578,41 @@ trait RangeOps extends Equal {
...
+          f(readVar(i))
+          i += 1
+        })
+
+      case Adapter.g.Def("range_until_step", List(x0:Backend.Exp, x1:Backend.Exp, x2:Backend.Exp)) =>
+        val i = var_new(Wrap[Int](x0))
+        __whileDo(ordering_lt(readVar(i), Wrap[Int](x1)), {
+          f(readVar(i))
+          i += Wrap[Int](x2)
+        })
```
feiwang3311, line 603: add support for `until` with `step`

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -47,9 +47,202 @@ trait CudaOps extends Base with SizeTOps with StackArrayOps with CLibs with Cuda
...
   // need to register the headers
   registerHeader("<cuda_header.h>")
+
+  override def remap(m: Manifest[_]): String = m.runtimeClass.getName match {
+    case s: String if s.endsWith("Dim3") => "dim3"
+    case _ => super.remap(m)
+  }
+
+  override def shallow(n: Node): Unit = n match {
+    case n @ Node(s,"@", (f:Backend.Exp)::args, _) if ((graphCache(f.asInstanceOf[Sym]) match {
```
feiwang3311, line 234: @Kraks we cannot use `Adapter.g.Def` in code gen because it is reset to `null` before  code gen

we can use `graphCache` though, which is a copy of `Adapter.g.graphCache`

src/out/thirdparty/cudacuda_global_fill.check.cu
```diff
@@ -0,0 +1,39 @@
...
+#include <stdint.h>
+#include <stdbool.h>
+/************* Functions **************/
+__global__ void x2(int* x3, int x4, int x5) {
+  int x6 = gridDim.x * blockDim.x;
+  int x7 = threadIdx.x + blockIdx.x * blockDim.x;
+  while (x7 < x5) {
+    x3[x7] = x4;
+    x7 = x7 + x6;
+  }
```
feiwang3311, line 17: we can generate CUDA kernel functions like this :)

src/out/thirdparty/cudacuda_global_fill.check.cu
```diff
@@ -0,0 +1,39 @@
...
+}
+/**************** Snippet ****************/
+void Snippet(int x0) {
+  int* x1 = (int*)malloc(0 * sizeof(int));
+  CUDA_CALL(cudaMalloc(&x1, (size_t)(5 * sizeof(int))));
+  x2<<<dim3(28, 1, 1), dim3(512, 1, 1)>>>(x1, 3, 5);
```
feiwang3311, line 23: we can use cuda kernel functions like this :)

src/main/scala/lms/core/codegen/C_codegen.scala
```diff
@@ -127,10 +127,11 @@ class ExtendedCCodeGen extends CompactCodeGen with ExtendedCodeGen {
     case _ => 20
   }
 
-  def emitFunction(name: String, body: Block) = {
+  def emitFunction(name: String, body: Block, decorator: String = "") = {
     val res = body.res
     val args = body.in
-    val ret = s"${remap(typeBlockRes(res))} "
+    val ret = if (decorator == "") s"${remap(typeBlockRes(res))} "
+              else s"$decorator ${remap(typeBlockRes(res))} "
```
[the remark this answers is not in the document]
feiwang3311: there is the difference of a space after decorator

src/main/scala/lms/core/stub.scala
```diff
@@ -259,15 +259,16 @@ trait Base extends EmbeddedControls with OverloadHack with lms.util.ClosureCompa
...
     val can = canonicalize(f)
     Adapter.funTable.find(_._2 == can) match {
       case Some((funSym, _)) => funSym
       case _ =>
         val fn = Backend.Sym(Adapter.g.fresh)
         Adapter.funTable = (fn, can)::Adapter.funTable
         val block = Adapter.g.reify(arity, gf)
-        val res = Adapter.g.reflect(fn, "λ", block, Backend.Const(0))()
+        val res = if (decorator == "") Adapter.g.reflect(fn, "λ", block, Backend.Const(0))()
+                  else Adapter.g.reflect(fn, "λ", block, Backend.Const(0), Backend.Const(decorator))()
```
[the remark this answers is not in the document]
feiwang3311: Yeah, I think you are right here :)

src/main/scala/lms/core/stub.scala
```diff
@@ -825,6 +840,16 @@ trait PrimitiveOps extends Base with OverloadHack {
...
 
+  // Generic?
+  implicit class GenericNegate[T:Numeric:Manifest](val self: Rep[T])(implicit __pos: SourceContext) {
+    def unary_- = Arithmetic.unary_-[T](self)
+
+    def +(rhs: Rep[T])(implicit __pos: SourceContext, __imp1: Overloaded1) = { Arithmetic.plus[T](self, rhs) }
+    def -(rhs: Rep[T])(implicit __pos: SourceContext, __imp1: Overloaded1) = { Arithmetic.minus[T](self, rhs) }
+    def *(rhs: Rep[T])(implicit __pos: SourceContext, __imp1: Overloaded1) = { Arithmetic.times[T](self, rhs) }
+    def /(rhs: Rep[T])(implicit __pos: SourceContext, __imp1: Overloaded1) = { Arithmetic.divide[T](self, rhs) }
+  }
```
feiwang3311, line 851: unfortunately this has duplications with what already in the code base

feiwang3311: Well, the even more unfortunate thing is that this duplication works in my terminal (maybe due to my higher java and sbt version) but doesn't pass gate test in here. 

How do we get this Numeric type working for Primitives such as + - * / ?

PR #43, opened 2020-08-01
[CUDA] more cuda kernel

feiwang3311:


src/main/resources/cuda_header.h
```diff
@@ -1,28 +1,103 @@
 #define NVIDIA_WARP_SIZE 32 // this is typically 32 (for incl. 1080ti s)
```
Kraks, line 1: It would be better if the header file is protected to prevent unexpected multiple includes.

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -210,17 +210,112 @@ trait CudaOps extends Base with RangeOps with SizeTOps with StackArrayOps with C
...
+      }
+    }
+
+  // cudaCap: cap the absolute value of `data`
+  def cudaCap[T:Numeric:Manifest](implicit __pos: SourceContext) = cudaGlobalFun {
+    (data: Rep[Array[T]], bound: Rep[T], size: Rep[Int]) =>
+      val stride = gridDimX * blockDimX
+      val tid = threadIdxX + blockIdxX * blockDimX
+      for (i <- tid.until(size, stride): Rep[Range]) {
+        __ifThenElse(data(i) > bound, {data(i) = bound}, {})
```
Kraks, line 243: use normal `if` syntax here?

src/main/scala/lms/core/stub.scala
```diff
@@ -840,6 +840,15 @@ trait PrimitiveOps extends Base with OverloadHack {
     def toLong(implicit __pos: SourceContext): Rep[Long] = cast_helper[Char,Long](self)
   }
 
+  // FIMXE(feiw) this overlaps with the implicit class of specific types
+  implicit class GenericMathOpsCls[T:Numeric:Manifest](val self: Rep[T])(implicit __pos: SourceContext) {
```
Kraks, line 844: So at later point, other classes for numeric syntax (like `PrimitiveMathOpsIntOpsCls` and `PrimitiveMathOpsLongOpsCls`) can be subsumed by `GenericMathOpsCls`? But it seems we still need to have a lot of boilerplate code when the operands have different types.

src/main/scala/lms/thirdparty/cuda/cuda.scala
feiwang3311: I tried. It works on ml00 machine but in GitHub Actions, it errors because I have to use `@virtualize`. (something like: the macro is defined and used at the same level)

src/main/scala/lms/core/stub.scala
feiwang3311: yeah, I don't know how to do so for different types.

src/main/resources/cuda_header.h
feiwang3311: Good point.

PR #46, opened 2020-08-06
[LMS-clean] support transformer

feiwang3311:
This PR is not supposed to be merged in as is, but I would like feed backs. 

I am trying to support adding transformation passes. 

See details in code diff and my comments.


src/main/scala/lms/core/stub.scala
```diff
@@ -24,6 +24,16 @@ object Adapter extends FrontEnd {
...
 
+  def genGraph1(m1: Manifest[_], m2: Manifest[_])(prog: Exp => Exp) = {
+    genGraphCommon(m1, m2)(g.reify(prog))
+  }
+  def genGraphCommon(m1: Manifest[_], m2: Manifest[_])(prog: => Block) = {
+    typeMap = new scala.collection.mutable.HashMap[lms.core.Backend.Exp, Manifest[_]]()
+    funTable = Nil
+    var g: Graph = time("staging") { program(prog) }
+    (typeMap, g)
+  }
```
feiwang3311, line 35: Adding support to get graph (instead of directly codegen) so that we can transform the graph

src/main/scala/lms/core/stub.scala
```diff
@@ -1257,6 +1267,29 @@ abstract class DslDriverCPP[A: Manifest, B: Manifest] extends DslDriverC[A, B] w
...
+    val source = new java.io.ByteArrayOutputStream()
+    val statics = time("codegen") {
+      codegen.typeMap = typeMap
+      codegen.stream = new java.io.PrintStream(source)
+      codegen.emitAll(g2, "Snippet")(manifest[A], manifest[B])
+      codegen.extractAllStatics.toList
+    }
+    (source.toString, statics)
+  }
+}
```
feiwang3311, line 1291: An example of compiler class that allows users to add several transformation passes before codegen :) 

I am adding 2 same transformations as a show case. Maybe should revise

Importantly, it is still not accurate because we are not updating the typeMap. It is OK for now because the transformer added is `identity transformation`. But this needs to be fixed too

src/main/scala/lms/core/traversal.scala
```diff
@@ -545,6 +545,75 @@ abstract class Transformer extends Traverser {
...
+  def transform(graph: Graph): Graph = {
+    Adapter.program(Adapter.g.reify { e =>
+      assert(graph.block.in.length == 1)
+      subst(graph.block.in(0)) = e
+      super.apply(graph);
+      transform(graph.block.res)
+    })
+  }
+}
+
```
feiwang3311, line 616: This is what I am not sure most. This `AdapterTransformer` class is the same as `Transformer` but just has the abstract `g` replaced with `Adapter.g`. I do this so that the transformer is using the same graphBuilder as all of our  XXXOps traits. This should allow us to use the front-end constructs in XXXOps traits in `Transformer` (instead of directly operating on graph). that is to say, we can do something like `printf` when transforming a Node(s, "show-tensor", _, _), for instance. 

How important do you guys think this is?

src/test/scala/lms/collection/test_arrays_trans.scala
```diff
@@ -0,0 +1,21 @@
...
+      @virtualize
+      def snippet(arg: Rep[Int]) = {
+        val x = NewArray[Int](10)
+        printf("%d\n", x(0))
+        x.free
+      }
+    }
+    check("free_1", driver.code, "c")
+  }
+}
```
feiwang3311, line 21: An old test using CompilerC

src/main/scala/lms/core/stub.scala
```diff
@@ -24,6 +24,16 @@ object Adapter extends FrontEnd {
   var typeMap: mutable.Map[lms.core.Backend.Exp, Manifest[_]] = _
```
Kraks, line 24: minor and unrelated: does typeMap have to be `var` and mutable map at the same time?

src/main/scala/lms/core/stub.scala
Kraks: seems `g` doesn't have to be `var` here. Also, I was thinking if we really want to have a mutable `typeMap` in Adapter, it is confusing that when should we reset it or pass it to codegen.

src/main/scala/lms/core/stub.scala
Kraks: AdapterTransformer could be just an object, so here we can just write `AdapterTransformer.transform(g)`.

Ideally, I think the composition of transformers is just function composition. But the typeMap effect breaks this compositionality so that we need to take additional care of typeMap.

src/main/scala/lms/core/stub.scala
feiwang3311: That is an interesting question. I think using `mutable typeMap` is probably faster than `immutable typeMap` when adding contents to the map? It is certainly temping to use `val` instead of `var` when the `typeMap` is mutable already. However, I am wondering if we want to reuse `Adapter` for transformation passes. If we do, we will have multiple graphs, each has a `typeMap` of its own. Then if `Adapter` is reused, we will need `Adapter.typeMap` to bind with a new hashMap. 

Or rather, just think about how one step of transformation can be done, when we want to use `Adapter.g.reflect` for the new (transformed) graph. We will have 2 typeMaps alive during the transformation. One is for the old graph, the other for the new graph. If both typeMaps are (or used to be) `Adapter.typeMap`, then I guess it has to be `var`?

src/main/scala/lms/core/stub.scala
feiwang3311: I don't think AdapterTransformer should be an object because I still want to extend it into other transformers such that different transformation passes can be built.

I do need to think about how typeMap works (not sure it is correct now). But as long as we reuse Adapter object, we will operation on typeMap. Somehow I need to save the old typeMap and reset it to a new one.

src/main/scala/lms/core/stub.scala
feiwang3311: I think it has to be `var`. 
I think it is better to be `mutable.Map` so that it is faster.

src/main/scala/lms/core/stub.scala
Kraks: I think instead of populating the side effects of updating `typeMap` and `funTable`, maybe we should encapsulate them into `GraphBuilder` (or `Graph`) and manipulate them through the interface from `GraphBuilder` (or `Graph`). In the current status, there is a chance that `typeMap`, `funTable` and graph `g` are mismatched.

Kraks: So that the `typeMap` can be accessed via `Adapter.g.typeMap`.

src/main/scala/lms/core/stub.scala
Kraks: I think if `typeMap` is already a mutable map, then it doesn't have to be var. In that case
```
typeMap = new scala.collection.mutable.HashMap[lms.core.Backend.Exp, Manifest[_]]()
```
can be just `typeMap.clear()`.

src/main/scala/lms/core/stub.scala
feiwang3311: Not sure. Tiark might want to have the `Graph` type-free, because it represents the simple core of LMS-IR. The type is more like a Meta-language checker and a code-gen guidance.

With that said, if we can NOT change the Graph or GraphBuilder, but extend them with `TypedGraph` and `TypedGraphBuilder`, I guess Tiark will think that is OK.

src/main/scala/lms/core/stub.scala
feiwang3311: No :) That is exactly why it has to be var. We need 2 typeMap being alive at the same time (one old, one new)
if we do `typeMap.clear()`, we don't have the old one any more.

Kraks: well, we can clone the map :) let me get a PR on that.

feiwang3311: but is `clone` and `clear` more expensive than moving a var?

Kraks: I guess atomically it is, but I don't think here it plays a big role on performance. See the proposed changes in my PR #47. It also uses `transform` instead of `map` when checking if two canonical representations exist in funTable.

src/main/scala/lms/core/stub.scala
```diff
@@ -37,8 +49,7 @@ object Adapter extends FrontEnd {
   def emitCommon(name: String, cg: ExtendedCodeGen, stream: java.io.PrintStream,
                  verbose: Boolean = false, alt: Boolean = false, eff: Boolean = false)
                 (m1: Manifest[_], m2: Manifest[_])(prog: => Block) = {
-    typeMap = new scala.collection.mutable.HashMap[lms.core.Backend.Exp, Manifest[_]]()
-    funTable = Nil
+    reset_state
```
Kraks, line 52:
```suggestion
    resetState
```

src/main/scala/lms/thirdparty/tensor1.scala
```diff
@@ -0,0 +1,78 @@
...
+        val p = print_tensor(shape, arr)(sc)
+        Unwrap(p)
+      case _ =>
+        System.out.println(graphCache(ts))
+        ???
+    }
+    case _ => super.transform(n)
+  }
+
+  def print_tensor(shape: Seq[Dim], arr: Backend.Exp, offset: Rep[Dim] = unit(0))(implicit pos: SourceContext): Rep[Unit] = shape match {
```
Kraks, line 61:
```suggestion
  def printTensor(shape: Seq[Dim], arr: Backend.Exp, offset: Rep[Dim] = unit(0))(implicit pos: SourceContext): Rep[Unit] = shape match {
```

src/main/scala/lms/core/stub.scala
Kraks: Having both untyped graph and TypedGraphsBuilder is okay to me, but if transformations can change the type, then it looks like we have to use a typed one.

feiwang3311: Even if transformation changes the type, it would generate a new typeMap I think. I don't think we `have to` use a typed graph, because a `typed graph` is "the same as" untyped graph + typeMap, right?

src/main/scala/lms/core/stub.scala
feiwang3311: reviewed that PR :)

src/main/scala/lms/core/traversal.scala
```diff
@@ -532,14 +535,35 @@ abstract class Transformer extends Traverser {
...
     // with traverser or with transform(Block)
+
+    // graphCache may be use during transformation to check the Node of a Sym
+    graphCache = graph.globalDefsCache
+
+    // Handling MetaData 1. save oldTypeMap/SourceMap first
+    val oldTypeMap = Adapter.typeMap
+    oldSourceMap = Adapter.sourceMap
+    // Handling MetaData 2. initialize MetaData as fresh, so the transformer might add new metadata entries
+    Adapter.typeMap = new scala.collection.mutable.HashMap[lms.core.Backend.Exp, Manifest[_]]()
```
Kraks, line 552:
```suggestion
    Adapter.typeMap = new scala.collection.mutable.HashMap[Backend.Exp, Manifest[_]]()
```

src/main/scala/lms/core/traversal.scala
```diff
@@ -532,14 +535,35 @@ abstract class Transformer extends Traverser {
...
+    Adapter.sourceMap = new scala.collection.mutable.HashMap[lms.core.Backend.Exp, SourceContext]()
```
Kraks, line 553:
```suggestion
    Adapter.sourceMap = new mutable.HashMap[Backend.Exp, SourceContext]()
```

src/main/scala/lms/core/traversal.scala
```diff
@@ -532,14 +535,35 @@ abstract class Transformer extends Traverser {
...
+
     val block = g.reify { e =>
       assert(graph.block.in.length == 1)
```
Kraks, line 556: what's the case that `block.in.length` is not 1?

src/main/scala/lms/core/traversal.scala
```diff
@@ -532,14 +535,35 @@ abstract class Transformer extends Traverser {
...
       subst(graph.block.in(0)) = e
       // subst(graph.block.ein) = g.curBlock.head // XXX
       super.apply(graph); transform(graph.block.res) }
```
Kraks, line 559:
```suggestion
      super.apply(graph)
      transform(graph.block.res) 
    }
```

src/main/scala/lms/transformation/tensor.md
```diff
@@ -0,0 +1,82 @@
...
+  } finally {g = null}
+}
+```
+
+The logic of this snippet should be clear.
+1. A `GraphBuilder` is constructed (the `mkGraphBuilder()` can be override by other classes).
+2. The `body` block is evaluted to `reflect` the nodes and `reify` the blocks into the `GraphBuilder`. Then a `Graph` object is returned containing all nodes, final block, and a hashmap from `Sym` to `Node` (for getting the definition of a `Sym`).
+3. Then the `g` is reset to null.
+
+In the more sofisticated frontend `stub.scala`, we use an `object Adapter` as the graph constructor.
```
Kraks, line 31:
```suggestion
In the more sophisticated frontend `stub.scala`, we use an `object Adapter` as the graph constructor.
```

src/main/scala/lms/transformation/tensor.md
```diff
@@ -0,0 +1,82 @@
...
+node construction.
+To offer a `Rep[T]` type for each `Backend.Exp`, the `Backend.Exp` are wrapped in a `Wrap[T]` class
+``` scala
+case class Wrap[+A:Manifest](x: lms.core.Backend.Exp) extends Exp[A] {
+   Adapter.typeMap(x) = manifest[A]
+}
+```
+And the mapping from `Backend.Exp` to `Manifest[_]` is kept in `Adapter.typeMap`.
+`Adapter.funTable` is another `Adapter` state that maps from frontend functions to their
+canonicanization. Note that after building the LMS IR in `Adapter`, the `typeMap` and `funTable`
```
Kraks, line 42:
```suggestion
canonicalized forms. Note that after building the LMS IR in `Adapter`, `typeMap` and `funTable`
```

src/main/scala/lms/transformation/tensor2.scala
```diff
@@ -0,0 +1,115 @@
+package lms
```
Kraks, line 1: why this file is called `tensor2.scala`? is there a `tensor1.scala` or something similar?

src/main/scala/lms/core/traversal.scala
feiwang3311: we do not handle that yet :)

src/main/scala/lms/transformation/tensor2.scala
feiwang3311: there was but it was not committed.
 I can rename this :)

PR #47, opened 2020-08-09
Refactor typeMap and funTable to immutable reference

Kraks:
The PR refactors `Adapter.typeMap` and `Adapter.funTable` to immutable reference (`val`) and adds several interfaces to effectully manipulate them. This should make latter refactor easier -- programming against interfaces instead of implementations.

### DIFF
--- a/src/main/scala/lms/core/stub.scala
+++ b/src/main/scala/lms/core/stub.scala
@@ -6,9 +6,10 @@ import lms.macros.EmbeddedControls
 
 import scala.collection.{mutable, immutable}
 
-import lms.core._
 import lms.util._
-import Backend._
+import lms.core._
+import lms.core.Backend._
+import lms.collection.mutable.ArrayOps
 
 import utils.time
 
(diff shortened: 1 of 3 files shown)


src/main/scala/lms/core/stub.scala
```diff
@@ -6,23 +6,34 @@ import lms.macros.EmbeddedControls
...
+  def copyTypeMap: mutable.Map[Backend.Exp, Manifest[_]] = typeMap.clone
+  def resetTypeMap: Unit = typeMap.clear
+  def updateTypeMap(e: Backend.Exp, m: Manifest[_]): Unit = typeMap(e) = m
+
+  private val funTable: mutable.ListBuffer[(Backend.Exp, Any)] = mutable.ListBuffer.empty
+  def copyFunTable: mutable.ListBuffer[(Backend.Exp, Any)] = funTable.clone
+  def resetFunTable: Unit = funTable.clear
+  def findFunction(can: Any): Option[(Backend.Exp, Any)] = funTable.find(_._2 == can)
+  def addFunction(f: Backend.Exp, can: Any): Unit = (f, can) +=: funTable
+  def simplifyFunTable(fn1: Backend.Exp, can1: Any): Unit =
```
feiwang3311, line 33: this is more like `updateFuntTable` I think

Kraks:
Postpone this PR until we have a better understanding of what metadata is needed and have measurements of performance.

PR #49, opened 2020-08-13
library path and compose paths

Shangyint:
Create the support for adding library path and concatenating paths for compiler command.

### DIFF
--- a/src/main/scala/lms/core/codegen/c_codegen.scala
+++ b/src/main/scala/lms/core/codegen/c_codegen.scala
@@ -260,6 +260,14 @@ class ExtendedCCodeGen extends CompactCodeGen with ExtendedCodeGen {
   val libraryFlags = mutable.HashSet[String]()
   def registerLibrary(nLibraries: String*) = libraryFlags ++= nLibraries.toSet
 
+  val libraryPaths = mutable.HashSet[String]()
+  def registerLibraryPath(paths: String*) = libraryPaths ++= paths.toSet
+
+  def composePaths(paths: mutable.HashSet[String], option: String): String = {
+    if (paths.isEmpty) ""
+    else s"$option ${paths.mkString(s" $option ")}"
+  }
+
   private val registeredFunctions = mutable.HashSet[String]()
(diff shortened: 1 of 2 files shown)


src/main/scala/lms/core/codegen/c_codegen.scala
```diff
@@ -260,6 +260,14 @@ class ExtendedCCodeGen extends CompactCodeGen with ExtendedCodeGen {
   val libraryFlags = mutable.HashSet[String]()
   def registerLibrary(nLibraries: String*) = libraryFlags ++= nLibraries.toSet
 
+  val libraryPaths = mutable.HashSet[String]()
+  def registerLibraryPath(paths: String*) = libraryPaths ++= paths.toSet
```
feiwang3311, line 264: is the libraryPaths used anywhere in compilation? Maybe it is used but not in this repo?

Shangyint: It would be used in the case where the linked library (static/dynamic) is not in the env's library paths. Not sure how often this would be encountered.

Kraks: It would be useful when we don't have permission to install libraries system-wide on some machines, under this case, we need to specify where are the .so files when compiling.

feiwang3311: make sense! but do we need to add something to the compilation cmd as well?

Shangyint: It would be something like 
`    val libraryPaths = codegen.composePaths(codegen.libraryPaths, "-L")`
Then add it to command.

src/main/scala/lms/core/codegen/c_codegen.scala
```diff
@@ -260,6 +260,14 @@ class ExtendedCCodeGen extends CompactCodeGen with ExtendedCodeGen {
...
+
+  def composePaths(paths: mutable.HashSet[String], option: String): String = {
```
Kraks, line 266: minor: `paths` can be a more general type, like `Iterable[String]`. I also would probably call this function `joinPaths`.

PR #54, opened 2020-08-27
[Transformation] better transformation

feiwang3311:


src/main/scala/lms/collection/ArrayOps.scala
```diff
@@ -8,7 +8,27 @@ import lms.core.virtualize
...
 
-trait ArrayOps { b: Base =>
+trait ArrayOps extends Base with PrimitiveOps {
+
+  case class ARRAY(x: Backend.Exp) {
+    def withSource(pos: SourceContext) = { Adapter.sourceMap(x) = pos; this }
+    def withEleType(m: Manifest[_]) = { Adapter.typeMap(x) = m; this }
+    def with_(pos: SourceContext, m: Manifest[_]) = withSource(pos).withEleType(m)
+
+    def p: SourceContext = Adapter.sourceMap.getOrElse(x, ???)
```
Kraks, line 18: Let's only use `???` for work in progress code, it is not intended to use `???` as a general exception. If there is a case that the program should abort, let it die with proper and meaningful error messages. In this case, if there is no default value at all, we can simply write `Adapter.sourceMap(x)`.

src/main/scala/lms/core/stub.scala
```diff
@@ -132,6 +132,61 @@ trait Base extends EmbeddedControls with OverloadHack with lms.util.ClosureCompa
...
+    def with_(pos: SourceContext, m: Manifest[_]) = withSource(pos).withType(m)
+
+    def p: SourceContext = Adapter.sourceMap.getOrElse(x, ???)
+    def t: Manifest[_] = Adapter.typeMap.getOrElse(x, ???)
+  }
+  def TOP(x: Backend.Exp, m: Manifest[_])(implicit __pos: SourceContext): TOP =
+    (new TOP(x)).with_(__pos, m)
+
+  case class UNIT(override val x: Backend.Exp)(implicit __pos: SourceContext) extends TOP(x) {
+    with_(__pos).asInstanceOf[UNIT]
```
Kraks, line 148: Why it needs an additional cast to `UINIT`?

src/main/scala/lms/core/stub.scala
```diff
@@ -132,6 +132,61 @@ trait Base extends EmbeddedControls with OverloadHack with lms.util.ClosureCompa
 
   implicit def unit[T:Manifest](x: T): Rep[T] = Wrap[T](Backend.Const(x))
 
+
+  class TOP(val x: Backend.Exp) {
```
Kraks, line 136: Any particular reason that `TOP` is not a case class?

src/main/scala/lms/core/stub.scala
```diff
@@ -132,6 +132,61 @@ trait Base extends EmbeddedControls with OverloadHack with lms.util.ClosureCompa
...
+    def withSource(pos: SourceContext) = { Adapter.sourceMap(x) = pos; this }
+    def withType(m: Manifest[_]) = { Adapter.typeMap(x) = m; this }
+    def with_(pos: SourceContext, m: Manifest[_]) = withSource(pos).withType(m)
```
Kraks, line 139: I would prefer to rename `with_` to `withSrcType` or something more semantic.

src/main/scala/lms/core/stub.scala
```diff
@@ -132,6 +132,61 @@ trait Base extends EmbeddedControls with OverloadHack with lms.util.ClosureCompa
...
+    val cBlock = Adapter.g.reify(c.x)
+    val bBlock = Adapter.g.reify({ b; Backend.Const(()) } )
+    // compute effect (cBlock bBlock)* cBlock
+    Adapter.g.reflectEffectSummary("W", cBlock,
+      bBlock)(Adapter.g.mergeEffKeys(cBlock, bBlock))
+  }
+
+
   // Functions
   def unwrapFun[A:Manifest,B:Manifest](f: Rep[A] => Rep[B]) =
```
Kraks, line 191:
```suggestion
  def unwrapFun[A:Manifest,B:Manifest](f: Rep[A] => Rep[B]): Backend.Exp => Backend.Exp =
```

src/main/scala/lms/core/stub.scala
```diff
@@ -132,6 +132,61 @@ trait Base extends EmbeddedControls with OverloadHack with lms.util.ClosureCompa
 
   implicit def unit[T:Manifest](x: T): Rep[T] = Wrap[T](Backend.Const(x))
 
+
+  class TOP(val x: Backend.Exp) {
+    def withSource(pos: SourceContext) = { Adapter.sourceMap(x) = pos; this }
```
Kraks, line 137: There is some redundancy of `withSource`, `withType`, and `with_`, but none of them override the other. it would be good to abstract the code a little bit using self type or traditional OO technique (such as F-bounded polymorphism) to express the invariant, like `withSource` should return the same type of this class, then it seems we can get rid of some of the castings.

src/main/scala/lms/core/stub.scala
```diff
@@ -132,6 +132,61 @@ trait Base extends EmbeddedControls with OverloadHack with lms.util.ClosureCompa
...
+  case class BOOL(override val x: Backend.Exp)(implicit __pos: SourceContext) extends TOP(x) {
+    with_(__pos).asInstanceOf[BOOL]
+    def withType: BOOL = { Adapter.typeMap(x) = manifest[Boolean]; this }
+    def with_(pos: SourceContext): BOOL =
+      withSource(pos).asInstanceOf[BOOL].withType
+
+    def unary_! = BOOL(Adapter.g.reflect("!",x))
+  }
+
+  case class VAR(x: Backend.Exp) {
```
Kraks, line 163: Why `VAR` does not extend `TOP`?

src/main/scala/lms/core/stub.scala
```diff
@@ -748,6 +803,46 @@ trait LiftPrimitives {
...
+  case class INT(override val x: Backend.Exp)(implicit __pos: SourceContext) extends NUM(x) {
+    with_(__pos, manifest[Int])
+    // def +(y: INT): INT = INT(Adapter.g.reflect("+", x, y.x))
+    // def -(y: INT): INT = INT(Adapter.g.reflect("-", x, y.x))
+    // def *(y: INT): INT = INT(Adapter.g.reflect("*", x, y.x))
+    // def /(y: INT): INT = INT(Adapter.g.reflect("/", x, y.x))
+  }
+  def INT(i: Int)(implicit __pos: SourceContext): INT = INT(Backend.Const(i))
+
+  def NUM(x: Backend.Exp, m: Manifest[_])(implicit __pos: SourceContext): NUM = {
```
Kraks, line 819:
```suggestion
  def NUM[T: Manifest](x: Backend.Exp)(implicit __pos: SourceContext): NUM = {
```
and `with_` can take `pos` and `m` as implicit parameters.

feiwang3311: The all CAPS classes and def constructors are supposed to be typeless (they don't take type parameters), so that the transformer function can use them. 
the `NUM[T: manifest]` would defeat the purpose.

src/main/scala/lms/core/stub.scala
feiwang3311: maybe have a chat about this? @Kraks

PR #56, opened 2020-09-05
Some sugar for emit and shallow

Kraks:
Defines a string interpolator `es` for `emit-and-shallow`, which turns an interpolated string such as `es"The quick brown $fox jumps over the lazy $dog"` into `emit("The quick brown "); shallow(fox); emit(" jumps over the lazy "); shallow(dog)`.

There is also an `esln` that prints a newline at the end.

### DIFF
--- a/src/main/scala/lms/core/codegen.scala
+++ b/src/main/scala/lms/core/codegen.scala
@@ -142,10 +142,36 @@ abstract class CompactCodeGen extends CompactTraverser {
       super.traverseCompact(ns, y)
     }
   }
+
   // the `emit` and `emitln` are basic codegen units.
   // the actual desitination of them are determined by children classes.
   def emit(s: String): Unit
   def emitln(s: String = ""): Unit
+
+  implicit class EmitHelper(val sc: StringContext) {
+    /** `es` for `emit-and-shallow`, which turns an interpolated string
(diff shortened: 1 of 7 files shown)


src/main/scala/lms/core/codegen.scala
```diff
@@ -142,11 +142,37 @@ abstract class CompactCodeGen extends CompactTraverser {
...
+          case (i: Int) => emit(i.toString)
+        }
+        emit(strings.next)
+      }
+    }
+    def esln(args: Any*): Unit = {
+      es(args:_*)
+      emitln()
+    }
+  }
```
feiwang3311, line 174: brilliant!
can you add more to the comments, explaining how this `implicit class` works and what becomes the `args: Any*`?

feiwang3311 (approved):
LGTM

src/main/scala/lms/core/codegen.scala
Kraks: That's the standard way to customize string interpolation in Scala, explained here https://docs.scala-lang.org/overviews/core/string-interpolation.html. I can also add this link to the comment. Roughly speaking, the intertwined string is decoupled to a sequence of raw strings `sc.parts` and a sequence of non-string arguments `args`, then we alternately output the head of two sequences.

PR #57, opened 2020-09-06
[LMS] more tensor support

feiwang3311:


src/main/scala/lms/core/stub.scala
```diff
@@ -803,37 +808,43 @@ object PrimitiveTypeLess {
...
     def *(y: NUM)(implicit pos: SourceContext): NUM = {
-      assert(t == y.t)
+      assert(t == y.t, s"t ${t} is not the same as y.t ${y.t} in NUM * method")
       NUM(Adapter.g.reflect("*", x, y.x), t)
     }
     def /(y: NUM)(implicit pos: SourceContext): NUM = {
-      assert(t == y.t)
+      assert(t == y.t, s"t ${t} is not the same as y.t ${y.t} in NUM / method")
       NUM(Adapter.g.reflect("/", x, y.x),t)
     }
```
feiwang3311, line 825: these methods are still bad because if I have a (INT + INT) I get a NUM (instead of INT) as result. How to fix?

src/main/scala/lms/core/stub.scala
```diff
@@ -803,37 +808,43 @@ object PrimitiveTypeLess {
 
   class NUM(override val x: Backend.Exp) extends TOP(x) {
     def +(y: NUM)(implicit pos: SourceContext): NUM = {
-      assert(t == y.t)
+      assert(t == y.t, s"t ${t} is not the same as y.t ${y.t} in NUM + method")
```
Kraks, line 811:
```suggestion
      assert(t == y.t, s"t ${t} is not the same as y.t ${y.t}")
```
When the assert fails, the call stack (as well as the last method) will be printed on the screen.

src/main/scala/lms/thirdparty/cpu_array_ops/ArrayCPUOps.scala
```diff
@@ -67,6 +110,41 @@ trait ArrayCPUOps extends Dsl with ArrayOps {
     }
   }
 
+  def array_minus[T:Numeric:Manifest](a: Rep[Array[T]], b: Rep[Array[T]], res: Rep[Array[T]], size: Int)(implicit __pos: SourceContext) = {
```
Kraks, line 113: Would `size` be `Rep[Int]`? If `size` is `Int`, the loop doesn't have to iterate over a `Rep[Range]`.

src/main/scala/lms/thirdparty/cpu_array_ops/ArrayCPUOps.scala
```diff
@@ -39,6 +41,47 @@ object ArrayCPUTypeLess extends Dsl with ArrayOps {
...
+  def ARRAY_DIV(a: ARRAY, b: ARRAY, res: ARRAY, size: INT)(implicit _pos: SourceContext) = {
+    for (i <- (0 until Wrap[Int](size.x)): Rep[Range]) {
+      val index = INT(Unwrap(i))
+      res(index) = a(index) / b(index)
+      ()
+    }
+  }
+
+  def ARRAY_VVDOT(a: ARRAY, b: ARRAY, res: ARRAY, size: INT)(implicit __pos: SourceContext) = {
+    // FIXME(feiw) do we need to initialize res with 0?
```
Kraks, line 69: I'd think the responsibility of initializing `res` is the caller's.

src/main/scala/lms/thirdparty/cpu_array_ops/ArrayCPUOps.scala
```diff
@@ -39,6 +41,47 @@ object ArrayCPUTypeLess extends Dsl with ArrayOps {
     }
   }
 
+  def ARRAY_MINUS(a: ARRAY, b: ARRAY, res: ARRAY, size: INT)(implicit _pos: SourceContext) = {
```
Kraks, line 44: There is some boilerplate code among `ARRAY_MINUS`, `ARRAY_MULT`, `ARRAY_DIV`, etc., maybe  we need to abstract over them later.

src/main/scala/lms/thirdparty/cpu_array_ops/cblas.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  class CBLAS_TRANSPOSE(override val x: Backend.Exp) extends TOP(x)
+  def CBLAS_TRANSPOSE(x: TOP): CBLAS_TRANSPOSE = new CBLAS_TRANSPOSE(x.x)
+  def NO_TRANS(implicit __pos: SourceContext): CBLAS_TRANSPOSE = CBLAS_TRANSPOSE(CMACRO("CblasNoTrans", manifest[Any]))
+  def YES_TRANS(implicit __pos: SourceContext): CBLAS_TRANSPOSE = CBLAS_TRANSPOSE(CMACRO("CblasTrans", manifest[Any]))
+
+  def CBLAS_SGEMV(layout: CBLAS_LAYOUT, transA: CBLAS_TRANSPOSE, m: INT, n: INT,
+      alpha: FLOAT, a: ARRAY, lda: INT, x: ARRAY, incX: INT, beta: FLOAT, y: ARRAY, incY: INT) = {
+    // this function only works for float arrays I think
+    assert(a.et == manifest[Float] && x.et == manifest[Float] && y.et == manifest[Float])
```
Kraks, line 33:
```suggestion
    require(a.et == manifest[Float] && x.et == manifest[Float] && y.et == manifest[Float])
```

src/main/scala/lms/transformation/devicedTensor.scala
```diff
@@ -0,0 +1,214 @@
...
+   *    even across graphs).
+   *  Another Solution (as implemented now) is to add a FLAG to the `class TENSOR` to indicate which set of Metadata should it use.
+   *    The FLAG is called `old`, which, if true, indicates that the `def et` (and other functions) should use the old set of Metadata.
+   *    How do we set the value of `old`? We use this convention.
+   *    1. We only use the `unsafe` constructor (new TENSOR(x)) in transformer for the old graph. So the `class TENSOR` has
+   *       the `old` FLAG default to true.
+   *    2. We always use the `safe` constructor (def TENSOR(..)) in non-transformer cases. It sets `old` to false.
+   *    3. We add another `safe` constructor (def tensor(a: Rep[Tensor[T]])) in typed constructor that always sets `old` to false.
+   */
+  class TENSOR(override val x: Backend.Exp, val old: Boolean = false) extends TOP(x) {
```
Kraks, line 72:
```suggestion
  class TENSOR(override val x: Backend.Exp, val useOldMetadata: Boolean = false) extends TOP(x) {
```

src/main/scala/lms/transformation/tensor.scala
```diff
@@ -40,17 +39,19 @@ object FixedSizeTensorTypeLess extends Base with PrimitiveOps with ArrayOps {
...
-  case class TENSOR(x: E) {
-    def withSource(pos: SourceContext) = { Adapter.sourceMap(x) = pos; this }
-    def withEleType(m: Manifest[_]) = { Adapter.typeMap(x) = m; this }
-    def withSrcType(pos: SourceContext, m: Manifest[_]) = withSource(pos).withEleType(m)
+  def TENSOR(shape: Seq[Int], array: ARRAY)(implicit __pos: SourceContext): TENSOR =
+    (new TENSOR(Adapter.g.reflectRead("tensor", C(shape), array.x)(array.x))).withSrcType(__pos, array.et)
 
-    def p: SourceContext = Adapter.sourceMap(x)
-    def et: Manifest[_] = Adapter.typeMap(x)
+  class TENSOR(override val x: Backend.Exp, val old: Boolean = true) extends TOP(x) {
```
Kraks, line 45: Similarly, `old` here.

src/main/scala/lms/core/stub.scala
```diff
@@ -803,37 +808,43 @@ object PrimitiveTypeLess {
...
 
   class INT(override val x: Backend.Exp) extends NUM(x) {
     this.withType(manifest[Int])
   }
   def INT(x: Backend.Exp)(implicit __pos: SourceContext): INT = (new INT(x)).withSource(__pos)
-  def INT(i: Int)(implicit __pos: SourceContext): INT = (new INT(Backend.Const(i))).withSource(__pos)
+  implicit def INT(i: Int)(implicit __pos: SourceContext): INT = (new INT(Backend.Const(i))).withSource(__pos)
+  def INT(x: TOP): INT = (new INT(x.x))
+
+  class FLOAT(override val x: Backend.Exp) extends NUM(x) {
```
Kraks, line 843: I wonder what's the datatype mapped to GPU? Is it still (and must be) float32 or it doesn't matter what we declared in frontend?

feiwang3311 (commented):
fixed :)

PR #58, opened 2020-09-06
[LMS clean] fix capture

feiwang3311:


src/main/scala/lms/core/codegen/cpp_codegen.scala
```diff
@@ -12,60 +12,62 @@ import lms.core.stub._
...
-  object Capture {
-    def apply(s: String) = Seq(CapAll(s))
-    def CopyExcept[T: Manifest](s: Rep[T]*) = Seq(CapAll("=")) ++ s.map(RefOne(_))
-    def RefExcept[T: Manifest](s: Rep[T]*) = Seq(CapAll("&")) ++ s.map(CopyOne(_))
-  }
   case class CapAll(s: String) extends Capture
   case class CopyOne[T: Manifest](s: Rep[T]) extends Capture
   case class RefOne[T: Manifest](s: Rep[T]) extends Capture
-  implicit def liftCapture(s:String) = Capture(s)
+
```
Kraks, line 18: It seems that `CapAll`, `CopyOne` and `RefOne` are only used when constructing `Captures`, and get immediately unwrapped when passed to `__fun`. The construction of `CopyExcept` and `RefExcept` adds a layer over `s`, but in codegen, it depends on the content of sentinel string (= or &). In other words, they (`CapAll` etc) don't have a semantic meaning in codegen. If we want to simplify things, I'd like to remove those but keeping the same interface provided in the object `Captures`.

feiwang3311: Good catch! fixed.

Kraks (approved):

PR #59, opened 2020-09-07
[TypeLessFrontend] fix case-only difference

feiwang3311:


src/main/scala/lms/core/stub.scala
```diff
@@ -791,13 +797,9 @@ trait LiftPrimitives {
...
 }
 
-/**
- * This file is extremely boilerplate. In fact, most of the code here is copied from a
- * Forge-generated file. We need a static version since Delite (and other projects) depend
- * on it without using Forge.
- */
-trait PrimitiveOps extends Base with OverloadHack {
 
+object PrimitiveTypeLess extends Base {
```
feiwang3311, line 801: this extends Base can be removed

src/main/scala/lms/transformation/tensor.scala
```diff
@@ -27,7 +27,10 @@ import Backend._
  * the typeless frontend is more friendly to IR transformations.
  * This is similar to how MLIR is typeless and has been easy to run transformations.
  */
-trait FixedSizeTensorFrontEnd extends Base with PrimitiveOps with ArrayOps {
+object FixedSizeTensorTypeLess extends Base with PrimitiveOps with ArrayOps {
```
feiwang3311, line 30: this extends Base ... cannot be removed, because in this typeless frontend, I actually used typed frontends :(  
not nice but this is with minimal effort

src/main/scala/lms/core/stub.scala
Kraks: indeed, if Rep etc. are not used in PrimitiveTypeLess

Kraks (approved):

src/main/scala/lms/transformation/tensor.scala
```diff
@@ -67,7 +70,11 @@ trait FixedSizeTensorFrontEnd extends Base with PrimitiveOps with ArrayOps {
   def TENSOR(shape: Seq[Int], array: ARRAY)(implicit __pos: SourceContext): TENSOR = {
     TENSOR(Adapter.g.reflect("tensor", C(shape), array.x)).withSrcType(__pos, array.et)
   }
+}
 
+trait FixedSizeTensorOps extends Base with PrimitiveOps with ArrayOps {
+  import ArrayTypeLess._
+  import FixedSizeTensorTypeLess._
```
feiwang3311, line 77: these 2 imports are expected because we like the idea of `typed frontend shallowly wraps typeless frontend` to reuse the graph construction logic

src/main/scala/lms/transformation/tensor2arrayCPU.scala
```diff
@@ -0,0 +1,80 @@
...
+import Backend._
+
+// lower Tensor computations to Array computations
+abstract class TensorLoweringCPU extends Transformer {
+
+  import BaseTypeLess._
+  import PrimitiveTypeLess._
+  import ArrayTypeLess._
+  import ArrayCPUTypeLess._
+  import FixedSizeTensorTypeLess._
```
feiwang3311, line 21: transformer now uses typeless frontends only

PR #61, opened 2020-09-16
[LSM-clean] simplify struct

feiwang3311:


src/main/scala/lms/thirdparty/c_lib_utils.scala
```diff
@@ -50,13 +50,13 @@ trait CCodeGenLibStruct extends ExtendedCCodeGen {
   }
 
   override def traverse(n: Node): Unit = n match {
-    case Node(s, "lib-struct", List(Const(m: Manifest[_])), _) =>
-      emit(s"${remap(m)} ")
+    case Node(s, "lib-struct", List(Const(m: String)), _) =>
+      emit(s"$m ")
       shallow(s)
       emitln(";")
```
Kraks, line 56:
```suggestion
      esln"$m $s;"
```
:)

src/main/scala/lms/thirdparty/c_lib_utils.scala
```diff
@@ -27,8 +27,8 @@ trait CCodeGenCMacro extends ExtendedCCodeGen {
 }
 
 trait LibStruct { b : Base =>
-  def newStruct[T: Manifest]: Rep[T] = {
-    Wrap[T](Adapter.g.reflectUnsafe("lib-struct", Backend.Const(manifest[T])))
+  def newStruct[T: Manifest](type_name: String): Rep[T] = {
```
Kraks, line 30:
```suggestion
  def newStruct[T: Manifest](typeName: String): Rep[T] = {
```

https://docs.scala-lang.org/style/naming-conventions.html

Kraks:
Except for those minor suggestions, I'm okay with this change, but I wonder what's the rationale behind it? It seems to be more flexible (and more unsafe) to control the generated type, and we don't need to provide the type mapping in the `remap` function?

feiwang3311:
> Except for those minor suggestions, I'm okay with this change, but I wonder what's the rationale behind it? It seems to be more flexible (and more unsafe) to control the generated type, and we don't need to provide the type mapping in the `remap` function?

The rationale is that if we only need to `declare a C struct` for a library, this allows us to only add code at one place. 
typeMap is needed only if the type is also used independently (such as in a function parameter).

Kraks (approved):
LGTM

PR #64, opened 2020-10-06
[AIRCoP] dim name support

feiwang3311:


src/main/scala/lms/transformation/util/utils.scala
```diff
@@ -0,0 +1,47 @@
+package lms.transformation.util
+
+import scala.collection.immutable._
+
+trait DataStructure {
```
feiwang3311, line 5: this is a Union Find data structure

Kraks: what's the purpose of this union find data structure?

feiwang3311: it is used to give Dim names (named dimension), which have to be unified based on operation features.

remember that in AIRCoP, the dims are named? However, if we don't expect the user to provide the names correctly, we have to unify the names based on operation features. Then we would need a union find data structure

PR #65, opened 2020-10-20
[ThirdParty] random data

feiwang3311:


src/main/scala/lms/core/stub.scala
```diff
@@ -806,6 +806,11 @@ trait LiftPrimitives {
 object PrimitiveTypeLess {
   import BaseTypeLess._
 
+  def CAST_HELPER(x: TOP, mX: Manifest[_], mY: Manifest[_], pos: SourceContext)(implicit __pos: SourceContext): TOP = {
```
Kraks, line 809: - if `x.t` has to be the same manifest as `mX`, why `mX` has to be provided? 
- why there is an explicit `pos` and an implicit `pos`?
- also wondering if this can be implemented in a more idiomatic way, eg, in the `class TOP` we define a method `castTo`:
```
def castTo(mY: Manifest[_])(implicit pos: SourceContext) = { ... }
```
or
```
def castTo[Y: Manifest](implicit pos: SourceContext) = { val mY = manifest[T]; ... }
```

src/main/scala/lms/thirdparty/random_data.scala
```diff
@@ -0,0 +1,44 @@
...
+  def rand_max(implicit __pos: SourceContext): INT = INT(CMACRO("RAND_MAX", manifest[Int]))
+
+  def rand_float(implicit __pos: SourceContext): FLOAT = {
+    val p = CAST_HELPER((rand_int - rand_max / 2), manifest[Int], manifest[Float], __pos)
+    val q = CAST_HELPER(rand_max, manifest[Int], manifest[Float], __pos)
+    FLOAT(NUM(p) / NUM(q))
+  }
+
+  def random_value(m: Manifest[_])(implicit __pos: SourceContext): TOP = m match {
+    case ma if ma == manifest[Int] => rand_int
```
Kraks, line 30: suggest to directly use `if`/`else` instead of pattern matching

src/main/scala/lms/thirdparty/random_data.scala
```diff
@@ -0,0 +1,44 @@
...
+
+import lms.core._
+import lms.util._
+import lms.core.stub._
+import lms.core.Backend._
+import lms.core.virtualize
+import lms.core.utils.time
+import lms.macros.SourceContext
+
+object RandomDataTypeLess {
```
Kraks, line 11: It seems to be a good idea to also have support for `srand`, something translated to `srand((unsigned) time(0))`.

src/main/scala/lms/thirdparty/random_data.scala
```diff
@@ -0,0 +1,44 @@
...
+  import BaseTypeLess._
+  import PrimitiveTypeLess._
+  import CLibTypeLess._
+
+  // FIXME(feiw) add support for this:
+  /* Intializes random number generator */
+  // srand((unsigned) time(&t));
+
+  def rand_int(implicit __pos: SourceContext): INT = INT(CMACRO("rand()", manifest[Int]))
```
Kraks, line 20: Add a `rand_int` that accepts an upper bound:
```suggestion
  def rand_int(implicit __pos: SourceContext): INT = INT(CMACRO("rand()", manifest[Int]))
  def rand_int(ub: Int)(implicit __pos: SourceContext): INT = INT(CMACRO(s"(rand() % $ub) + 1", manifest[Int]))
```

src/main/scala/lms/thirdparty/random_data.scala
```diff
@@ -0,0 +1,44 @@
...
+    case ma if ma == manifest[Float] => rand_float
+    case ma => throw new Exception(s"manifest $ma is not yet handled in random_value function")
+  }
+}
+
+trait RandomDataOps extends Base {
+
+  import RandomDataTypeLess._
+
+  def randInt(implicit __pos: SourceContext): Rep[Int] = Wrap[Int](rand_int.x)
```
Kraks, line 40:
```suggestion
  def randInt(implicit __pos: SourceContext): Rep[Int] = Wrap[Int](rand_int.x)
  def randInt(ub: Int)(implicit __pos: SourceContext): Rep[Int] = Wrap[Int](rand_int(ub).x)
```

src/main/scala/lms/thirdparty/random_data.scala
feiwang3311: Added

src/main/scala/lms/thirdparty/random_data.scala
feiwang3311: Added

src/main/scala/lms/thirdparty/random_data.scala
feiwang3311: Added

Kraks (approved):
LGTM

PR #66, opened 2020-11-17
[Transformation] add facility for logging graphs

feiwang3311:


src/main/scala/lms/core/stub.scala
```diff
@@ -1388,19 +1388,22 @@ abstract class DslDriverCPP[A: Manifest, B: Manifest] extends DslDriverC[A, B] w
 abstract class CompilerC[A:Manifest, B:Manifest] extends DslDriverC[A, B] { q =>
 
   // get original graph
-  val graph = Adapter.genGraph1(manifest[A], manifest[B])(x => Unwrap(wrapper(Wrap[A](x))))
+  val initial_graph = Adapter.genGraph1(manifest[A], manifest[B])(x => Unwrap(wrapper(Wrap[A](x))))
 
   // run some transformation
-  def transform(graph: Graph): Graph = graph
+  def transform(graph: Graph): (List[String], Graph) = (List(graph.toString), graph)
```
Kraks, line 1394: Does it make sense to return a `List[Graph]` which contains all intermediate graphs instead of `(List[String], Graph)`? And the first/last one in the list can be the final/initial graph. I just felt it is not `transform`'s job to return several string representations of the graphs.

feiwang3311: That can be done

Kraks (approved):
LGTM

PR #71, opened 2020-12-08
cudnn and test_cudnn

luke-jiang:


src/main/scala/lms/thirdparty/cuda/cudnn.scala
```diff
@@ -0,0 +1,407 @@
...
+
+
+trait CUDNNOps extends CLibs with CudaOps {
+
+  /* LMS support for CUDNN library */
+  
+  class cudnnStatusT
+
+  class cudnnHandleT
+  def cudnnHandle: Rep[cudnnHandleT] = newStruct[cudnnHandleT]("cudnnHandle_t")
```
feiwang3311, line 35: we might also want to provide a `cached` cudnnHandle, so that one cudnnHandle is used throughout.

You can check how cublasHandle is done :)

luke-jiang: cached cudnnHandle added to cudnn.scaka

src/main/scala/lms/thirdparty/cuda/cudnn.scala
```diff
@@ -0,0 +1,545 @@
...
+trait CCodeGenCUDNN extends ExtendedCCodeGen {
+
+  registerHeader("\"cudnn_header.h\"")
+
+  override def remap(m: Manifest[_]) = m.runtimeClass.getName match {
+    case s: String if s.endsWith("$cudnnConvolutionFwdAlgoPerfT") => "cudnnConvolutionFwdAlgoPerf_t"
+    case s: String if s.endsWith("$cudnnConvolutionFwdAlgoT") => "cudnnConvolutionFwdAlgo_t"
+    case _ => super.remap(m)
+  }
+}
```
feiwang3311, line 545: Nit: maybe add a new line at the end of the files

src/test/scala/lms/thirdparty/test_cudnn.scala
```diff
@@ -0,0 +1,367 @@
...
+        cudnnCheck(cudnnDestroyTensorDescriptor(output_descriptor))
+        cudnnCheck(cudnnDestroyDropoutDescriptor(dropout_descriptor))
+
+        cudnnCheck(cudnnDestroy(cudnn))
+      }
+    }
+    // System.out.println(indent(driver.code))
+    check("cudnn_dropout", driver.code, "cu")
+  }
+}
```
feiwang3311, line 367: Nit: same here :)

feiwang3311 (approved):
Looks great!

src/main/scala/lms/thirdparty/cuda/cudnn.scala
feiwang3311: fixed

src/test/scala/lms/thirdparty/test_cudnn.scala
feiwang3311: fixed

PR #72, opened 2021-01-16
Cublas test

luke-jiang:
add three test cases for cublas, verifying cublasSgemm, cublasSgeam, cublasSgemv and cublasSdot.
add cublas_header.h


src/main/resources/headers/cublas_header.h
```diff
@@ -0,0 +1,16 @@
...
+  {                                                              \
+    cublasStatus_t err = (f);                                    \
+    if (err != CUBLAS_STATUS_SUCCESS) {                          \
+      fprintf(stderr, "CUBLAS error occurred\n");                \
+      exit(err);                                                 \
+    }                                                            \
+  }                                                              \
+
+
+#endif
```
feiwang3311, line 16: I know it sounds silly but GitHub doesn't like it if the end of files are not newline. 
When that happens, you see a circle with minus at the bottom of the file.

feiwang3311 (approved):
Looks great.
Just wondering if you tested the generated .cu files on ml servers.

src/main/resources/headers/cublas_header.h
luke-jiang: I checked on the server. 

I'll fix the format :)

PR #77, opened 2021-02-04
First Unary: negation

Harjas12:
- I currently do not have access to machine where I can run the negation.check.cu. If one of you could run it and verify that it is correct, I would appreciate that.
- This PR just contains the negation operator. If this is good without any issues, I will put a new PR with all the other unary OPs in one bigger PR.


src/main/scala/lms/transformation/mpi_nccl_transformation/distributedTensor2MPINCCLUnary.scala
```diff
@@ -0,0 +1,69 @@
...
+          op match {
+            case "tensor_negate" => gpu_neg_array(count, m, myNCCLRank, loaded_operand).x
+            case _ => throw new Exception(s"op $op is not unary op")
+          }
+        case SAnno(dim: Dim, devices: Seq[Device], _) => throw new Exception(s"TODO: not yet handling SAnno with AllReduce")
+        case a => throw new Exception(s"TODO: annotation $a is not yet handled")
+      }
+    case _ => super.transform(n)
+  }
+}
```
feiwang3311, line 69: GitHub wants you to add a new line here

Harjas12: added newline

PR #79, opened 2021-02-11
AIRCoP Unaries

Harjas12:
Implemented:
- [x] Invert
- [x] Transpose
- [x] Relu
- [x] Tanh

Compiles on machine:
- [x] Invert
- [x] Transpose
- [x] Relu
- [x] Tanh

Fix backward nodes
- [x] Invert
- [x] Tanh
- [x] Relu


src/main/scala/lms/transformation/distributed_tensor/distributedTensorUnary.scala
```diff
@@ -45,6 +63,30 @@ trait FixedSizeDistributedTensorUnaryTypeLess extends FixedSizeDistributedTensor
         (() => {
           Accumulate(gradMap(a), Negate(gradMap(s), anno), anno); ()
         }) +=: backwardNodes
+     
+    case Node(s, "tensor_invert", tt::Backend.Const(anno:Anno)::(a:Backend.Sym)::_, _) =>
+        implicit val pos = Adapter.oldSourceMap(s)
+        forwardNodes += node
+
+        (() => {
+          Accumulate(gradMap(a), Invert(gradMap(s), anno), anno); ()
```
feiwang3311, line 72: I think we need to fix the computation for backward pass, and actually some of them are a bit tricky.
for instance, the `invert` is like y = f(x) where f(x) = 1/x. Then if y is used to compute z by z = g(y), we can apply the chain-rule, such that we know dz/dx = dz/dy * dy/dx.
as a result, the operations in `backwardNodes` should be dz/dy * dy/dx. We know that `dz/dy` is gradMap(s). Then the computation needed is to multiply it with `dy/dx`, in this case, 1/x^2.

So I think the code here should be
val square = Mul(transform(a), transform(a), anno)
val grad = Div(gradMap(s), square, anno)
Accumulate(gradMap(a), grad, anno); ()

it is also possible for us to enclose these computations to a new node (maybe called InvertGrad)

src/main/scala/lms/transformation/distributed_tensor/distributedTensorUnary.scala
```diff
@@ -45,6 +63,30 @@ trait FixedSizeDistributedTensorUnaryTypeLess extends FixedSizeDistributedTensor
...
+        }) +=: backwardNodes
+    
+    case Node(s, "tensor_tanh", tt::Backend.Const(anno:Anno)::(a:Backend.Sym)::_, _) =>
+        implicit val pos = Adapter.oldSourceMap(s)
+        forwardNodes += node
+
+        (() => {
+          Accumulate(gradMap(a), Tanh(gradMap(s), anno), anno); ()
```
feiwang3311, line 80: the derivative of tanh is a little more complex, but you can refer to this https://blogs.cuit.columbia.edu/zp2130/derivative_of_tanh_function/

Actually the computations here are fairly complex, so that maybe we can define a TanhGrad node that just do that. TanhGrad is likely a binary op.

src/main/scala/lms/transformation/distributed_tensor/distributedTensorUnary.scala
```diff
@@ -45,6 +63,30 @@ trait FixedSizeDistributedTensorUnaryTypeLess extends FixedSizeDistributedTensor
...
+        }) +=: backwardNodes
+    
+    case Node(s, "tensor_relu", tt::Backend.Const(anno:Anno)::(a:Backend.Sym)::_, _) =>
+        implicit val pos = Adapter.oldSourceMap(s)
+        forwardNodes += node
+
+        (() => {
+          Accumulate(gradMap(a), Relu(gradMap(s), anno), anno); ()
```
feiwang3311, line 88: we don't need to worry about derivative of Relu. However, Relu actually behaves as a mask (only positive values are used after Relu). As a result, the gradient (dz/dy) should also be filtered by the same mask before adding to (dz/dx).
so I think we might need a ReluGrad node that takes the input or output of Relu (as a mask), the gradMap(s), and return a masked gradMap(s). such as
Accumulate(gradMap(a), ReluGrad(transform(a or s), gradMap(s), anno), anno); ()
where the computation of ReluGrad is similar to Relu, but it masks the gradMap(s) argument.

src/main/scala/lms/transformation/tensor/tensor.scala
```diff
@@ -100,6 +100,14 @@ object FixedSizeTensorTypeLess extends Base with PrimitiveOps with ArrayOps {
...
       (new TENSOR(Adapter.g.reflect("tensor_dot", C(res_shape), x, y.x))).withSrcType(__pos, et)
     }
+
+    def neg()(implicit __pos: SourceContext): TENSOR = {
+      (new TENSOR(Adapter.g.reflect("tensor_negate", C(shape), x))).withSrcType(__pos, et)
+    }
+
+    def inv()(implicit __pos: SourceContext): TENSOR = {
+      (new TENSOR(Adapter.g.reflect("tensor_invert", C(shape), x))).withSrcType(__pos, et)
+    }
```
feiwang3311, line 110: we don't need to update this file :) 
the /tensor.scala is deprecated

feiwang3311 (commented):
actually I didn't realize that the backward computation for these forward computations are fair complex. They have to be fixed for the test to work. If that make the PR too big, maybe we can split this PR to 3 PRs (one for each pair of unaryOp and unaryGradOp).

src/main/scala/lms/transformation/distributed_tensor/distributedTensorUnary.scala
Harjas12: I have gone ahead and made your suggested changes

src/main/scala/lms/transformation/distributed_tensor/distributedTensorUnary.scala
Harjas12: I looked at the tanh derivative. I get why it needs to be different, but why would it need to be a binary op? The derivate still only takes 1 variable

Harjas12: I updated how the tanh gradient is calculated. I didn't make its own nodes because I didn't understand the value in doing so. I can't imagine how we would ever reuse the tanh derivative.

src/main/scala/lms/transformation/distributed_tensor/distributedTensorUnary.scala
Harjas12: I added the ReluGrad node. I am not sure if I fully understood how the gradient of Relu was supposed to work so I did a little extra research and found this: https://blog.paperspace.com/vanishing-gradients-activation-function/.
I essentially implemented ReluGrad as x > 0 => 1, x < 0 => 0

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -177,6 +180,26 @@ object CUDATypeLess extends Dsl with StackArrayOps with CLibs with CudaFunction
...
+      val size = (new INT(xn(2))).withSrcType(__pos, manifest[Int])
+
+      val x = blockIdxX * tileDim + threadIdxX
+      val y = blockIdxY * tileDim + threadIdxY
+      val width = gridDimX * tileDim
+
+      val start = Backend.Const(0)
+      for (i <- range_until_step(Wrap[Int](start), Wrap[Int](tileDim.x), Wrap[Int](blockSize.x))) {
+        val index = INT((y + INT(Unwrap(i))) * width + x)
+        out(index) = in(index); ()
```
feiwang3311, line 197: this seems a bit strange.
does this make `out` the same as `in`?

src/main/scala/lms/transformation/distributed_tensor/distributedTensorUnary.scala
```diff
@@ -45,6 +69,33 @@ trait FixedSizeDistributedTensorUnaryTypeLess extends FixedSizeDistributedTensor
...
         }) +=: backwardNodes
+     
+    case Node(s, "tensor_invert", tt::Backend.Const(anno:Anno)::(a:Backend.Sym)::_, _) =>
+        implicit val pos = Adapter.oldSourceMap(s)
+        forwardNodes += node
+
+        (() => {
+          val a_tensor = new TENSOR(transform(a))
+          val square = Mul(a_tensor, a_tensor, anno)
+          val grad = Div(gradMap(a), square, anno)
```
feiwang3311, line 80: should this be `gradMap(s)` instead of `gradMap(a)`?
also there seems to be missing a negation?
if y = x^-1, then dy/dx = -1x^-2

Another thing to take into account is that this implementation will likely contain 3 CUDA kernels (the mul, the div, the neg). It would be interesting if we can use 1 CUDA kernel that does all three things. It can be done either via a later transformation/optimization pass, or by defining a `Invert_grad` kernel that will later lower to a CUDA kernel that does this three things.

src/main/scala/lms/transformation/distributed_tensor/distributedTensorUnary.scala
```diff
@@ -45,6 +69,33 @@ trait FixedSizeDistributedTensorUnaryTypeLess extends FixedSizeDistributedTensor
...
+          Accumulate(gradMap(a), grad, anno); ()
+        }) +=: backwardNodes
+    
+    case Node(s, "tensor_tanh", tt::Backend.Const(anno:Anno)::(a:Backend.Sym)::_, _) =>
+        implicit val pos = Adapter.oldSourceMap(s)
+        forwardNodes += node
+
+        (() => {
+          Accumulate(gradMap(a), TanhGrad(gradMap(s), new TENSOR(transform(a)), anno), anno); ()
```
feiwang3311, line 89: TanhGrad might want to take `transform(s)` as input instead? 
otherwise it has to compute tanh again IIRC.

src/main/scala/lms/thirdparty/cuda/cuda.scala
Harjas12: oh I accidentally committed the wrong code when I was testing around. I updated the code based on the naive transpose from: https://developer.nvidia.com/blog/efficient-matrix-transpose-cuda-cc/

src/main/scala/lms/transformation/distributed_tensor/distributedTensorUnary.scala
Harjas12: done.

src/main/scala/lms/thirdparty/cuda/cuda.scala
Harjas12: I committed an experiment rather than I was intending on doing here by accident. I have gone ahead and corrected it. My source: https://developer.nvidia.com/blog/efficient-matrix-transpose-cuda-cc/

src/main/scala/lms/transformation/distributed_tensor/distributedTensorUnary.scala
Harjas12: Done.

src/main/scala/lms/core/stub.scala
```diff
@@ -904,6 +904,34 @@ object PrimitiveTypeLess {
...
+      BOOL(Adapter.g.reflect(">", x, y.x))
+    }
+
+    def tanh()(implicit pos: SourceContext): NUM = {
+      NUM(Adapter.g.reflect("tanh", x), t)
+    }
+
+    def tanh_grad(y: NUM, m: Manifest[_])(implicit pos: SourceContext): NUM = {
+      this * (NUM_ONE(m) - y * y)
+    }
```
feiwang3311, line 919: I think we probably don't need the `m: Manifest[_]` here, since you can get the manifest via `y.t`

src/main/scala/lms/transformation/distributed_tensor/distributedTensorUnary.scala
```diff
@@ -24,9 +24,27 @@ trait FixedSizeDistributedTensorUnaryTypeLess extends FixedSizeDistributedTensor
...
+    val res_tt = tensor.resultType
+    (new TENSOR(Adapter.g.reflectRead("tensor_relu", C(res_tt), C(anno), tensor.x)(tensor.x)).withSrcType(__pos, tensor.et))
+  }
+
   override def mergable_dims(node: Node) = node match {
     case Node(s, "tensor_transpose", _, _) => List()
     case Node(s, "tensor_negate", _, _) => List()
+    case Node(s, "tensor_invert", _, _) => List()
+    case Node(s, "tensor_tanh", _, _) => List()
+    case Node(s, "tensor_relu", _, _) => List()
```
feiwang3311, line 47: maybe we can build a local variable called `unaryops`, and here, use the `if unaryops.contains(op)` instead

feiwang3311 (approved):
just 2 tiny comments

src/main/scala/lms/transformation/distributed_tensor/distributedTensorUnary.scala
Harjas12: done.

src/main/scala/lms/core/stub.scala
Harjas12: done.

PR #82, opened 2021-02-25
AIRCoP softmax

luke-jiang:
Implement AIRCoP softmax and activation

- [x] how to deal with softmax algorithm and mode? Should user be able to specify that?
- [x] `getConvDescriptor` doesn't add back to map
- [x] `finalize_cudnn` doesn't emit typeless DESTROY calls
- [x] create test case for activation
- [x] change `overwriteCheckFiles` to false before merge

### DIFF
--- a/src/main/scala/lms/transformation/distributed_tensor/distributedTensorConv.scala
+++ b/src/main/scala/lms/transformation/distributed_tensor/distributedTensorConv.scala
@@ -71,6 +71,22 @@ trait FixedSizeDistributedTensorConvTypeLess extends FixedSizeDistributedTensorM
 
     TensorType(output_shape, input.et, anno)
   }
+  
+  def SoftmaxForward(input: TENSOR, params: SoftmaxParam, anno: Anno, __pos: SourceContext): TENSOR = {
+    val SoftmaxParam(alpha, beta) = params
+
+    val res_tt = input.resultType
+    (new TENSOR(Adapter.g.reflectRead("tensor_softmax", C(res_tt), C(anno), input.x, 
+      C(params))(input.x)).withSrcType(__pos, input.et))
+  }
+
+  def SoftmaxBackward(output: TENSOR, doutput: TENSOR, params: SoftmaxParam, anno: Anno, __pos: SourceContext): TENSOR = {
+    val SoftmaxParam(alpha, beta) = params
+
+    val res_tt = doutput.resultType
+    (new TENSOR(Adapter.g.reflectRead("tensor_softmax_bwd", C(res_tt), C(anno), output.x, doutput.x, 
+      C(params))(output.x, doutput.x)).withSrcType(__pos, doutput.et))
+  }
 
   override def mergable_dims(node: Node) = node match {
     // constraints:
@@ -85,6 +101,8 @@ trait FixedSizeDistributedTensorConvTypeLess extends FixedSizeDistributedTensorM
       val filterCout  = filter_type(CUDNN_C_OUT).dim
       val filterCin   = filter_type(CUDNN_C_IN).dim
       List((inputC, filterCin), (outputC, filterCout))
+
+    case Node(s, op, _, _) if op == "tensor_softmax" => List()
 
     case _ => super.mergable_dims(node)
   }
@@ -111,6 +129,16 @@ trait FixedSizeDistributedTensorConvTypeLess extends FixedSizeDistributedTensorM
           val b_grad = ConvBackwardFilter(x, y, gradMap(s), params, anno, pos)
           Accumulate(gradMap(b), b_grad, anno); ()
         }) +=: backwardNodes
+      
+      case Node(s, "tensor_softmax", tt::Backend.Const(anno:Anno)::(a:Backend.Sym)::Backend.Const(params:SoftmaxParam)::_, _) =>
+        implicit val pos = Adapter.oldSourceMap(s)
+        forwardNodes += node
+        (() => {
+            val x = new TENSOR(transform(a))
+            val grad = SoftmaxBackward(x, gradMap(s), params, anno, pos)
+            Accumulate(gradMap(a), grad, anno); ()
+           ()
+        }) +=: backwardNodes
 
       case _ => super.aircopCollect(node, forwardNodes, weightNodes, backwardNodes, gradMap, momentumMap, transform)
     }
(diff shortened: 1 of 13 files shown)


src/out/transformer/distributed_tensor/softmax.check.cu
```diff
@@ -0,0 +1,191 @@
...
+    int x52 = 0;
+    while (x52 != 18) {
+      printf("%f ", x51[x52]);
+      x52 = x52 + 1;
+    }
+    printf("\n");
+    // end copying GPU array x298 to CPU and print for size 18 and type Float
+  }
+  printf("compile");
+  CUDNNCHECK(cudnnDestroy(x8));
```
luke-jiang, line 177: should also include CUDNNCHECK(cudnnDestroyTensorDescriptor(x39)), since x39 is a descriptor created, and should be destroyed by finalize_cudnn

src/main/scala/lms/transformation/distributed_tensor/distributedTensorConv.scala
```diff
@@ -71,6 +71,22 @@ trait FixedSizeDistributedTensorConvTypeLess extends FixedSizeDistributedTensorM
 
     TensorType(output_shape, input.et, anno)
   }
+  
+  def SoftmaxForward(input: TENSOR, params: SoftmaxParam, anno: Anno, __pos: SourceContext): TENSOR = {
+    val SoftmaxParam(alpha, beta) = params
```
feiwang3311, line 76: this line seems not necessary

src/main/scala/lms/transformation/distributed_tensor/distributedTensorConv.scala
```diff
@@ -71,6 +71,22 @@ trait FixedSizeDistributedTensorConvTypeLess extends FixedSizeDistributedTensorM
...
+
+    val res_tt = input.resultType
+    (new TENSOR(Adapter.g.reflectRead("tensor_softmax", C(res_tt), C(anno), input.x, 
+      C(params))(input.x)).withSrcType(__pos, input.et))
+  }
+
+  def SoftmaxBackward(output: TENSOR, doutput: TENSOR, params: SoftmaxParam, anno: Anno, __pos: SourceContext): TENSOR = {
+    val SoftmaxParam(alpha, beta) = params
```
feiwang3311, line 84: this line seems not necessary

src/main/scala/lms/transformation/distributed_tensor/distributedTensorConv.scala
```diff
@@ -86,6 +102,8 @@ trait FixedSizeDistributedTensorConvTypeLess extends FixedSizeDistributedTensorM
       val filterCin   = filter_type(CUDNN_C_IN).dim
       List((inputC, filterCin), (outputC, filterCout))
 
+    case Node(s, op, _, _) if op == "tensor_softmax" => List()
```
feiwang3311, line 105: would be simpler to just do `case Node(s, "tensor_softmax", _, _) => List()`

src/main/scala/lms/transformation/distributed_tensor/distributedTensorConv.scala
```diff
@@ -111,6 +129,16 @@ trait FixedSizeDistributedTensorConvTypeLess extends FixedSizeDistributedTensorM
...
         }) +=: backwardNodes
+      
+      case Node(s, "tensor_softmax", tt::Backend.Const(anno:Anno)::(a:Backend.Sym)::Backend.Const(params:SoftmaxParam)::_, _) =>
+        implicit val pos = Adapter.oldSourceMap(s)
+        forwardNodes += node
+        (() => {
+            val x = new TENSOR(transform(a))
+            val grad = SoftmaxBackward(x, gradMap(s), params, anno, pos)
+            Accumulate(gradMap(a), grad, anno); ()
+           ()
```
feiwang3311, line 140: I am not sure about this one:
should it be
`val grad = SoftmaxBackward(x, gradMap(s), params, anno, pos)`
or
`val grad = SoftmaxBackward(new TENSOR(transform(s)), gradMap(s), params, anno, pos)`?

The API https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnSoftmaxBackward seems to be using `yDesc` and `y`, but I am not certain.

luke-jiang: I think it's the second one, but I have an error when running on the server.

src/main/scala/lms/transformation/distributed_tensor/distributedTensorConv.scala
luke-jiang: fixed in 232d098

src/main/scala/lms/transformation/distributed_tensor/distributedTensorConv.scala
luke-jiang: fixed in 232d098

src/main/scala/lms/transformation/distributed_tensor/distributedTensorConv.scala
luke-jiang: fixed in 232d098

src/out/transformer/distributed_tensor/softmax.check.cu
```diff
@@ -0,0 +1,201 @@
...
+  if (x46) {
+    int x51 = x1;
+    int x52 = 0;
+    while (x52 != x51) {
+      int x53 = x52;
+      NCCLCHECK(ncclRecv(x47 + x53 * 9, (size_t)18, ncclFloat32, x53, x5, x6));
+      x52 = x52 + 1;
+    }
+  }
+  NCCLCHECK(ncclGroupEnd());
```
luke-jiang, line 172: NCCL failure softmax.check.cu:172 'unhandled cuda error'

src/test/scala/lms/utils.scala
```diff
@@ -10,7 +10,7 @@ trait LibSuite extends FunSuite {
 }
 
 trait TutorialFunSuite extends LibSuite {
-  val overwriteCheckFiles = false // should be false; temporary set to true only to simplify development
+  val overwriteCheckFiles = true // should be false; temporary set to true only to simplify development
```
feiwang3311, line 13: don't forget to revert this change.

feiwang3311 (approved):
LGTM Thanks!

src/test/scala/lms/utils.scala
luke-jiang: fixed in 5deaa7c

PR #83, opened 2021-03-01
[AIRCOP] some fix on cudnn ops

luke-jiang:
- [x] change `traverse` in `DistributeTensor2MPI_NCCLAnalysis` to use a list of cudnn ops
- [x] change names of `gpu_array1_...` functions
- [ ] remove fwd-bwd hashmaps
- [x] implement more unary ops for activation. If user doesn't specify parameter, use unary. Otherwise, use cudnn activation.
- [x] implement pooling
- [x] implement pooling user API
- [x] fix unused getoutputdim in conv (current implement is not correct because var_read is not used)
- [x] fix alpha/beta pointer bug in sgemm
- [ ] more cudnn ops?
- [x] cleanup
- [x] change `overwriteCheckFiles` to false before merge

### DIFF
--- a/src/test/scala/lms/utils.scala
+++ b/src/test/scala/lms/utils.scala
@@ -10,7 +10,7 @@ trait LibSuite extends FunSuite {
 }
 
 trait TutorialFunSuite extends LibSuite {
-  val overwriteCheckFiles = true // should be false; temporary set to true only to simplify development
+  val overwriteCheckFiles = false // should be false; temporary set to true only to simplify development
 
   val prefix = "src/out/"
   val under: String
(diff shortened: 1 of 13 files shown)


src/test/scala/lms/utils.scala
```diff
@@ -10,7 +10,7 @@ trait LibSuite extends FunSuite {
 }
 
 trait TutorialFunSuite extends LibSuite {
-  val overwriteCheckFiles = false // should be false; temporary set to true only to simplify development
+  val overwriteCheckFiles = true // should be false; temporary set to true only to simplify development
```
feiwang3311, line 13: revert this :)

luke-jiang: fixed in 9ff6254

Harjas12 (approved):

feiwang3311 (approved):

PR #88, opened 2021-03-19
generate basic cuda kernel

luke-jiang:


src/test/scala/lms/thirdparty/test_cuda.scala
```diff
@@ -164,5 +164,71 @@ class CudaTest extends TutorialFunSuite {
...
+        val dynamicReverse = cudaGlobalFun[Array[Int], Int, Unit]((d, n) => {
+          val s = NewDynSharedArray[Int]
+          val t = threadIdxX
+          val tr = n - t - 1
+          s(t) = d(t)
+          cudaSyncThreads
+          d(t) = s(tr)
+        })
+
+        val tmp = NewShared2dArray[Int](2,2)
```
luke-jiang, line 191: NewShared2dArray is not transformed

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -405,6 +408,18 @@ trait CudaOps extends Dsl with StackArrayOps with SizeTOps with CLibs with CudaF
     addr
   }
 
+  def NewSharedArray[T:Manifest](x: Rep[Int]): Rep[Array[T]] = {
+    Wrap[Array[T]](Adapter.g.reflectMutable("NewSharedArray", Unwrap(x)))
+  }
+
+  def NewShared2dArray[T:Manifest](x: Rep[Int], y: Rep[Int]): Rep[Array[Array[T]]] = {
```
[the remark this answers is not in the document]
luke-jiang: fixed in e407dce

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -405,6 +408,22 @@ trait CudaOps extends Dsl with StackArrayOps with SizeTOps with CLibs with CudaF
...
+    Wrap[Array[T]](Adapter.g.reflectMutable("NewSharedArray", Unwrap(x)))
+  }
+
+  def NewSharedArray[T:Manifest](x: Rep[Int], y: Rep[Int]): Rep[Array[Array[T]]] = {
+    Wrap[Array[Array[T]]](Adapter.g.reflectMutable("NewSharedArray", Unwrap(x), Unwrap(y)))
+  }
+
+  def NewSharedArray[T:Manifest](x: Rep[Int], y: Rep[Int], z: Rep[Int]): Rep[Array[Array[Array[T]]]] = {
+    Wrap[Array[Array[Array[T]]]](Adapter.g.reflectMutable("NewSharedArray", Unwrap(x), Unwrap(y), Unwrap(z)))
+  }
```
feiwang3311, line 421: Just gave more thought to it and I think all three functions here should take `Int` typed parameter instead of `Rep[Int]`, since the parameter should be some const value in generated code. Do you agree?

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -597,16 +616,23 @@ trait CudaOps extends Dsl with StackArrayOps with SizeTOps with CLibs with CudaF
...
     Wrap[(A,Dim3,Dim3)=>B](__topFun(f, 1, xn => Unwrap(f(Wrap[A](xn(0)))), "__global__"))
 
-  def cudaGlobalFun[A:Manifest,B:Manifest,C:Manifest](f: (Rep[A], Rep[B]) => Rep[C])=
+  def cudaGlobalFun[A:Manifest,B:Manifest,C:Manifest](f: (Rep[A], Rep[B]) => Rep[C]) =
     Wrap[(A,B,Dim3,Dim3)=>C](__topFun(f, 2, xn => Unwrap(f(Wrap[A](xn(0)), Wrap[B](xn(1)))), "__global__"))
-
+  
+  def cudaGlobalDynamicFun[A:Manifest,B:Manifest,C:Manifest](f: (Rep[A], Rep[B]) => Rep[C]) =
+    Wrap[(A,B,Dim3,Dim3,Dim1)=>C](__topFun(f, 2, xn => Unwrap(f(Wrap[A](xn(0)), Wrap[B](xn(1)))), "__global__"))
+
```
feiwang3311, line 635: probably need this function for `f` of other arity too

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -729,6 +755,11 @@ trait CCodeGenCudaOps extends CCodeGenSizeTOps with CudaCodeGenLibFunction with
   // need to register the headers
   registerHeader("\"cuda_header.h\"")
 
+  override def mayInline(n: Node): Boolean = n match {
+    case Node(s, "NewSharedArray", _, _) => false
```
feiwang3311, line 759: need NewDynSharedArray too I think

feiwang3311 (approved):
Looks Great! Just a few comments :)

src/test/scala/lms/thirdparty/test_cuda.scala
```diff
@@ -164,5 +164,83 @@ class CudaTest extends TutorialFunSuite {
...
+        })
+
+        val dynamicReverse = cudaGlobalDynamicFun[Array[Int], Int, Unit]((d, n) => {
+          val s = NewDynSharedArray[Int]
+          val t = threadIdxX
+          val tr = n - t - 1
+          s(t) = d(t)
+          cudaSyncThreads
+          d(t) = s(tr)
+        })
```
feiwang3311, line 189: let's also move these 2 functions to cuda.scala (as a library that every other test can use if it wants to)

PR #92, opened 2021-03-28
Cuda Transpose Kernel

Harjas12:
- [x] write kernel
- [ ] write transpose test

### DIFF
--- a/src/main/scala/lms/thirdparty/cuda/cuda.scala
+++ b/src/main/scala/lms/thirdparty/cuda/cuda.scala
@@ -814,14 +814,14 @@ trait CudaOps extends Dsl with StackArrayOps with SizeTOps with CLibs with CudaF
     }
 
   def cudaTranspose[N:Numeric:Manifest](implicit __pos: SourceContext) = cudaGlobalFun {
-    (in: Rep[Array[N]], out: Rep[Array[N]]) => 
+    (in: Rep[Array[N]], out: Rep[Array[N]]) =>
       val tile = NewSharedArray[N](tileDim, tileDim + 1)
       
       val x = blockIdxX * tileDim + threadIdxX
       val y = blockIdxY * tileDim + threadIdxY
       val width = gridDimX * tileDim
 
-      for(i <- (0 until (tileDim, blockRows)): Rep[Range]) {
+      for (i <- (0 until (tileDim, blockRows)): Rep[Range]) {
         tile(threadIdxY + i)(threadIdxX) = in((y + i) * width + x)
       }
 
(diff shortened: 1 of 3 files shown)


luke-jiang (approved):

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -700,7 +701,7 @@ trait CudaOps extends Dsl with StackArrayOps with SizeTOps with CLibs with CudaF
   def threadIdxX: Rep[Int] = cmacro[Int]("threadIdx.x")
   def threadIdxY: Rep[Int] = cmacro[Int]("threadIdx.y")
   def threadIdxZ: Rep[Int] = cmacro[Int]("threadIdx.z")
-
+  def blockRows: Rep[Int] = cmacro[Int]("BLOCK_ROWS")
```
feiwang3311, line 704: I think `BLOCK_ROWS` is a local macro defined in their code, not a macro we can just use :)

Harjas12: added it as a constant into our code

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -661,10 +661,12 @@ trait CudaOps extends Dsl with StackArrayOps with SizeTOps with CLibs with CudaF
 
 
   // Some global values for our GPU
+  val tileDim = 32
```
feiwang3311, line 664: Nit: I probably want to move the declaration of `tileDim` and `blockRows` right before the `cudaTranspose` function, since they are not global values for GPU, but rather variables used by the transpose kernel

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -812,6 +813,28 @@ trait CudaOps extends Dsl with StackArrayOps with SizeTOps with CLibs with CudaF
       }
     }
 
+  def cudaTranspose[N:Numeric:Manifest](implicit __pos: SourceContext) = cudaGlobalFun {
+    (in: Rep[Array[N]], out: Rep[Array[N]]) =>
```
feiwang3311, line 817: Nit: we would want some comments here via `generate_comment(xxx)` function. Those comments will be generated in the generated code, explaining what the parameters are and how the caller is supposed to call it (what is the Dim3 used for instance)

feiwang3311:
since we have merged cuda transpose kernel, should this PR be closed?

PR #93, opened 2021-03-30
[AIRCoP] masked fill

luke-jiang:
implement masked fill according to the code in Lantern: https://github.com/feiwang3311/Lantern/blob/2dd710d3b773c75674d10384b3bb62b1f46a153a/src/main/cpp/headers/cublas_header.h#L1189

- [x] implement masked_fill forward and backward kernels
- [x] implement test case for fwd pass with python equivalent
- [x] test on server


src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -799,6 +814,68 @@ trait CudaOps extends Dsl with StackArrayOps with SizeTOps with CLibs with CudaF
       val i = blockIdxX * blockDimX * threadIdxX
       __ifThenElse(i < size, {y(i) = a * x(i) + y(i)}, {})
   }
+
+  def maskedFill[N:Numeric:Manifest](ijSwapped: Boolean)(implicit __pos: SourceContext) = cudaGlobalFun {
+    (in: Rep[Array[N]], out: Rep[Array[N]], mask: Rep[Array[Int]], value: Rep[N], 
+    dim0_shape: Rep[Int], dim1_shape: Rep[Int], dim0_stride: Rep[Int], dim1_stride: Rep[Int],
+    input_size: Rep[Int]) => {
+
```
feiwang3311, line 822: let's add some comments here to explain the parameters and how the caller should call it (what is the dim3 to use)

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -799,6 +814,68 @@ trait CudaOps extends Dsl with StackArrayOps with SizeTOps with CLibs with CudaF
...
+      })
+    }
+  }
+
+  
+  def maskedFillGrad[N:Numeric:Manifest](ijSwapped: Boolean)(implicit __pos: SourceContext) = cudaGlobalFun {
+    (y_d: Rep[Array[N]], x_d: Rep[Array[N]], mask: Rep[Array[Int]],
+    dim0_shape: Rep[Int], dim1_shape: Rep[Int], dim0_stride: Rep[Int], dim1_stride: Rep[Int],
+    input_size: Rep[Int]) => {
+
```
feiwang3311, line 854: comments here to 
via `generate_comment` function

src/test/scala/lms/thirdparty/test_cuda.scala
```diff
@@ -286,5 +287,67 @@ class CudaTest extends TutorialFunSuite {
...
+        cudaCall(cudaMemcpyOfT[Float](dinput, dinput_d, n, device2host))
+
+        printf("masked fill input gradient:\n")
+        for (i <- (0 until n): Rep[Range]) {
+          printf("%f,", dinput(i))
+        }
+        printf("\n")
+      }
+    }
+    check("kernel_masked_fill", driver.code, "cu")
```
feiwang3311, line 350: if you want to pull in this PR https://github.com/TiarkRompf/lms-clean/pull/95
we can start adding pytorch golden files, data, and check in the test :)

supunab (approved):

src/test/scala/lms/thirdparty/test_cuda.scala
```diff
@@ -342,5 +362,96 @@ class CudaTest extends TutorialFunSuite {
...
+
+        val mask = NewArray[Int](n)
+        scanFile[Int]("golden/maskedFill/mask.data", mask, n)
+        val cuda_mask = cudaMalloc2[Int](n)
+        cudaCall(cudaMemcpyOfT[Int](cuda_mask, mask, n, host2device))
+
+        val output = NewArray[Float](n)
+        val cuda_output = cudaMalloc2[Float](n)
+        val maskedFillKernel = maskedFill[Float](true)
+        maskedFillKernel(cuda_input, cuda_output, cuda_mask, maskValue, d0, d1, d0, 1, n, dim3((n + 511)/512), dim3(512))
```
luke-jiang, line 449: I don't think the kernel is called properly :(

supunab: Yeah, I think the way it's called is slightly inaccurate. Below is a summary of the semantics of maskedFill (we can add this the Scala code as a comment).

```
/*
    Semantics of maskedFill:
    Let's take a n-d tensor input = tensor.randn(10, 2, 5, 8)
    Now let's say we want to apply a mask to the above tensor for dimension 1 and dimension 2;
    Therefore, the shape of the mask should be (2, 5) in this case.
    (if the mask was (5, 2) and needed to apply to dimension 2 and dimension 1, we mark ijSwapped = true (because 2>1))

    We name the two dimensions as dim0 and dim1.
    We need the shape of the two dims (dim0_shape = 2, dim1_shape = 5)
    We also need the strides for the two dims (dim0_stride = 5*8, dim1_stride = 8)
  */
```

Now, in this test case scenario, `ijSwapped` must be `false` because we apply the mask to dimension 0 and 1. And the rest of the params should be as follows;

```
d0_shape = 8
d1_shape = 8
d0_stride = 8 // = d0_shape
d1_stride = 1
```

Can you give this a try and see if it matched the PT output?

luke-jiang: It worked! Thank you so much for the detailed summary. I understand it now :)

supunab (approved):

PR #97, opened 2021-04-03
Clean-ups and Renamings

luke-jiang:
- Move old cudnn `check.cu` files into a new `cuda/cudnn` directory
- Rename test cases in `test_cuda.scala` s.t. all kernel test cases end up with `_kernel`
- Add `softmax.check.cu`
- Move kernel implementations into `CudaLibs`


supunab (approved):

feiwang3311 (approved):
just small comments

Harjas12 (approved):

PR #98, opened 2021-04-04
Generic Cuda Transpose Kernel

Harjas12:
converted cuda transpose kernel work with any arbitrary (n x m) matrix


src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -866,29 +866,33 @@ trait CudaOps extends Dsl with StackArrayOps with SizeTOps with CLibs with CudaF
...
       val tile = NewSharedArray[N](tileDim, tileDim + 1)
-      val x = blockIdxX * tileDim + threadIdxX
-      val y = blockIdxY * tileDim + threadIdxY
-      val width = gridDimX * tileDim
 
-      for(i <- (0 until (tileDim, blockRows)): Rep[Range]) {
-          tile(threadIdxY + i, threadIdxX) = in((y + i) * width + x)
+      for (i <- (0 until (tileDim, blockRows)): Rep[Range]) {
+        val x = blockIdxX * tileDim + threadIdxX
+        val y = blockIdxY * tileDim + threadIdxY + i
```
feiwang3311, line 878: could some of the computation be left before this for loop?

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -866,29 +866,33 @@ trait CudaOps extends Dsl with StackArrayOps with SizeTOps with CLibs with CudaF
...
 
       cudaSyncThreads
 
-      val row = blockIdxY * tileDim + threadIdxX
-      val col = blockIdxX * tileDim + threadIdxY
-
       for (i <- (0 until (tileDim, blockRows)): Rep[Range]) {
-          out((col + i) * width + row) = tile(threadIdxX, threadIdxY + i)
+        val x = blockIdxY * tileDim + threadIdxX
+        val y = blockIdxX * tileDim + threadIdxY + i
```
feiwang3311, line 889: same here

feiwang3311 (approved):
Looks Great! Just some nits

src/main/scala/lms/thirdparty/cuda/cuda.scala
Harjas12: done

src/main/scala/lms/thirdparty/cuda/cuda.scala
Harjas12: don

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -866,29 +866,36 @@ trait CudaOps extends Dsl with StackArrayOps with SizeTOps with CLibs with CudaF
   }
 
   def cudaTranspose[N:Numeric:Manifest](implicit __pos: SourceContext) = cudaGlobalFun {
-    (in: Rep[Array[N]], out: Rep[Array[N]]) =>
+    (in: Rep[Array[N]], out: Rep[Array[N]], n: Rep[Int], m: Rep[Int]) =>
```
Harjas12, line 869: @feiwang3311 Should I add tileDim and blockRows as parameters to the kernel? The calling function and the kernel need to agree on these constants for it to work correctly, so it might reduce errors if they were parameters instead of assumed.

feiwang3311: That is reasonable. But i think asking the caller to provide `tileDim` and `blockRows` is risky too, because the caller might use unreasonable `tileDim` or `blockRows`? I think maybe the comments of this function should state clearly what is `tileDim` and `blockRows` (since they are constant for kernel writer) and how it should be used by the caller.

feiwang3311 (approved):

src/main/scala/lms/thirdparty/cuda/cuda.scala
Harjas12: I think that's a good way of handling it. I have gone ahead and added enough comments to the function that anyone should be able to figure out how to call it.

PR #101, opened 2021-04-12
[Cuda] split concat

luke-jiang:
- [ ] generate comment for general split/concat kernel
- [ ] change `overwriteCheckFiles` to false before merge


src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -1093,6 +1093,55 @@ trait CudaLibs extends CudaOps {
...
+  def cudaSplit3D0[N:Numeric:Manifest](implicit __pos: SourceContext) = cudaGlobalFun {
+    (in: Rep[Array[N]], out1: Rep[Array[N]], out2: Rep[Array[N]], d0: Rep[Int], split: Rep[Int], input_size: Rep[Int]) => {
+      generate_comment("this is cuda split kernel. It takes a 3D array and splits on the innermost dimension (dim2).")
+      generate_comment("in: input array")
+      generate_comment("out1: first output array")
+      generate_comment("out2: second output array")
+      generate_comment("d0: dim2 of the input array")
+      generate_comment("split: dim2 of the first output array")
+      generate_comment("call constraint: 0 < split < in.dim2 ")
+      generate_comment("call constraint: out1.dim2 + out2.dim2 = in.dim2 ")
```
feiwang3311, line 1106: I think using `out1` or `d0` in the `generate_comments` is not ideal. If you check the generated code, the parameters do not keep the scala name. So I think the best we can do is to use `arg0` `arg1` ..

In the generated comments, I think one important comment is how the kernel is called, i.e., what is the assumed `dim3(x,y,z)`.

luke-jiang: fixed in 1f17da3

src/out/thirdparty/cuda/concat2.check.cu
```diff
@@ -0,0 +1,74 @@
...
+  // arg5: product of other two dimensions (dim0 * dim1)
+  // call constraint: arg1 + arg3 = out.dim2
+  int x14 = blockIdx.x * blockDim.x + threadIdx.x;
+  int x15 = x9 + x11;
+  if (x14 < x13 * x15) {
+    int x16 = x14 % x15;
+    x12[x14] = x16 < x9 ? x8[x14 / x15 * x9 + x16] : x10[x14 / x15 * x11 + (x16 - x9)];
+  }
+}
+__global__ void x17(float* x18) {
```
luke-jiang, line 32: `x2` and `x4` are not in scope

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -1093,6 +1093,161 @@ trait CudaLibs extends CudaOps {
...
+          } else {
+            __ifThenElse(y < offs(t+1), { get_case(t) }, { make_case(t+1) })
+          }
+        }
+        make_case(0)
+      }, {})
+    }
+  }
+
+  def cuda3DConcat[N:Numeric:Manifest](n: Int, ds: List[Int], in: List[Rep[Array[N]]])(implicit __pos: SourceContext) = cudaGlobalFun {
```
luke-jiang, line 1222: passing `in` here may not be a good idea

supunab (approved):
Looks good to me!
(Double checked the tests by running on ML cluster)

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -1175,6 +1178,113 @@ trait CudaLibs extends CudaOps {
...
+      generate_comment(s"call constraint: in.size = $n")
+      generate_comment(s"call constraint: sum of in(i).size = out.size for i in [0, $n)")
+
+      val idx = blockIdxX * blockDimX + threadIdxX
+
+      __ifThenElse(idx < input_size, {
+        val x = idx / dimX
+        val y = idx % dimX
+
+        def get_case(t: Int) = {
```
luke-jiang, line 1258: maybe `get_case` and `make_case` can be changed to better names?

feiwang3311: I agree. let me fix them

PR #102, opened 2021-04-13
[AIRCoP] [CUDA] LogSoftmax and misc

supunab:
Adding support for logSoftmax (can be easily used to compute Cross Entropy Loss) and testing for both Softmax and LogSoftmax using PyTorch programs.


luke-jiang (approved):

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -1065,9 +1068,13 @@ trait CudaLibs extends CudaOps {
     cudaSyncThreads
   }
 
-  def softmax[N:Numeric:Manifest](log: Boolean)(implicit  __pos: SourceContext) = cudaGlobalDynamicFun {
+  def softmax[N:Numeric:Manifest](logSoftmax: Boolean)(implicit __pos: SourceContext) = cudaGlobalDynamicFun {
```
luke-jiang, line 1071: maybe use `cudaSoftmax`, `cudaSoftmaxGrad` to keep naming consistency with other kernels?

supunab: yep, makes sense! will update. Thanks!

PR #103, opened 2021-04-13
[CUDA] Generic Permute3D implementation

feiwang3311:


src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -429,10 +431,18 @@ trait CudaOps extends Dsl with StackArrayOps with SizeTOps with CLibs with CudaF
       val offset = skip * i + j
       Wrap[T](Adapter.g.reflectRead("array_get", Unwrap(x), Unwrap(offset))(Unwrap(x)))
     }
+    def apply(is: List[Rep[Int]])(implicit __pos: SourceContext): Rep[T] = {
```
luke-jiang, line 434: what is the benefit for using `List[Rep[Int]]` instead of using 2 parameters?

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -1142,6 +1162,221 @@ trait CudaLibs extends CudaOps {
...
+      for (i <- (0 until (tileDim, blockRows)): Rep[Range]) {
+        out_offset(blockIdxY, blockIdxX, threadIdxX, threadIdxY + i, tile(threadIdxX, threadIdxY + i))
+      }
+  }
+
+  // amazing helper function to get flatten offset of tensors
+  def flatten(shape: List[Rep[Int]], index: List[Rep[Int]])(implicit __pos: SourceContext): Rep[Int] = {
+    val a: Rep[Int] = (shape.tail zip index.init).foldLeft(unit(0)) {
+      case (z: Rep[Int], (shape: Rep[Int], index: Rep[Int])) => (index + z) * shape
+      }
```
luke-jiang, line 1212: This `}` maybe better inlined with `val` :)

src/main/scala/lms/thirdparty/cuda/cuda.scala
```diff
@@ -1142,6 +1162,221 @@ trait CudaLibs extends CudaOps {
...
+      generate_comment("caller must use <<<dim3(dimY, dimZ, 1), dim3(A, 1, 1)>>> where A < dimX")
+      generate_comment("each threadblock hands one dimX in coalease size of A, then we have dimZ x dimY threadblocks")
+      generate_comment("this kernel might be inefficient if the dimX is small. TODO")
+
+      // amazing helper function to get flatten offset of tensors
+      def flatten(shape: List[Rep[Int]], index: List[Rep[Int]]): Rep[Int] = {
+        val a: Rep[Int] = (shape.tail zip index.init).foldLeft(unit(0)) {
+          case (z: Rep[Int], (shape: Rep[Int], index: Rep[Int])) => (index + z) * shape
+          }
+        a + index.last
```
luke-jiang, line 1361: perhaps extra tab here?

luke-jiang (approved):

src/main/scala/lms/thirdparty/cuda/cuda.scala
feiwang3311: just convenient if the indices are already in a list :)

src/main/scala/lms/thirdparty/cuda/cuda.scala
feiwang3311: good point

PR #104, opened 2021-04-17
embedding grad

Harjas12:
- [x] embedding grad small
- [x] testcase: embedding grad small
- [ ] embedding grad large
- [ ] testcase: embedding grad large


supunab (approved):
LGMT

PR #110, opened 2021-07-21
updates on lms doc

luke-jiang:


YuyanBao:
In the file lms_code_walk.md: 
in high level -> at high level?

In the file traversal.md: 
This functionality is handled by the `Transformer` class, -> This is handled by the `Transformer` class,?
I am not sure about the phrasing of control dependencies. I thought there were (effect) dependencies based on the email during the POPL submission.
We define the bound variables of these new symbol -> We define the bound variables of these new symbols ?
For each node in `g`, get all bound variables that it depends on  -> For each node in `g`, get all bound variables it depends on ?
then schedule the node outside means that the node -> then scheduling the node outside means that the node?

doc/main/scala/lms/core/backend.md
```diff
@@ -1,45 +1,35 @@
...
 case class Const(x: Any) extends Exp // Constant
 case class Block(in: List[Sym], res: Exp, ein: Sym, eff: EffectSummary) extends Def
+
 case class Node(n: Sym, op: String, rhs: List[Def], eff: EffectSummary)
 ```
 
-From this class hierarchy, we can see that the basic unit of the LMS IR is `Exp`,
-which can be `Const(0)`, `Const("hello")`, or `Sym(2)`, which really means `x2` in the
-generated code. We can combine basic units into `Node`, such as
+`Def` represents a definition, which is used in the RHS of all nodes. A definition can be an `Exp` or a `Block`. `Exp` can be a `Sym` or a `Const`. `Sym` represents symbols in the IR identified by a unique integer. For example, `Sym(2)` represents `x2` in the generated code. `Const` represents immediate values, such as integers (e.g. `Const(0)`), strings (e.g. `Const(hello)`), etc.
```
Kraks, line 16:
```suggestion
`Def` represents a definition, which is used in the RHS of all nodes. A definition can be an `Exp` or a `Block`. `Exp` can be a `Sym` or a `Const`. `Sym` represents symbols in the IR identified by a unique integer. For example, `Sym(2)` represents `x2` in the generated code. `Const` represents immediate values, such as integers (e.g. `Const(0)`), strings (e.g. `Const("hello")`), etc.
```

doc/main/scala/lms/core/backend.md
```diff
@@ -1,45 +1,35 @@
...
 
+`Node` represents statements in the IR. The first field `n` of a node is a `Sym` corresponding to the LHS of a statement. The second field `op` is the operator of the statement. We use strings as operators since they permit easy extension of various kinds of nodes such as `-`, `print`, etc. The third field is the list of operands (RHS) of a node. Let's ignore the last field `EffectSummary` for now, which will be covered later. For example, the statement `val x3 = x2 + 1` would become
```
Kraks, line 18:
```suggestion
`Node` represents statements in the IR. The first field `n` of a node is a `Sym` corresponding to the identity of a statement. The second field `op` is the operator of the statement. We use strings as operators since they permit easy extension of various kinds of nodes such as `-`, `print`, etc. The third field is the list of operands (RHS) of a node. Let's ignore the last field `EffectSummary` for now, which will be covered later. For example, the statement `val x3 = x2 + 1` would become
```

doc/main/scala/lms/core/backend.md
```diff
@@ -57,38 +47,37 @@ for (i <- 0 until 100) {
...
+The criteria we use in LMS IR are *dependencies*. Simply speaking, dependencies capture information flow in the source program that we must respect. Given the order of nodes in the source program defined by the programmer, we can see that certain nodes can be reordered or moved in/out of scopes without changing the semantics of the program, but certainly, that's not true for all nodes. Dependency information tells us the order of nodes we must keep when we schedule nodes. There are two kinds of dependencies that we care about in LMS IR:
 1. the data dependency and
 2. the control dependencies.
 
-The data dependency is easier to capture. If a node uses an `Exp`, then the node
-depends on that `Exp`, and the node that generates that `Exp`. The data dependency
-is also captured explicitly by the graph. For instance, in code:
+The data dependency is straightforward. If a node uses an `Exp`, then the node
+depends on that `Exp` and the node that generates that `Exp`. Data dependency is captured explicitly by our definition of `Node`. If the `rhs` of node A contains node B, then A is data-dependent on B. For instance, in the following code snippet:
```
Kraks, line 61:
```suggestion
depends on that `Exp` and the node that generates that `Exp`. Data dependency is captured explicitly by our definition of `Node`. If the `rhs` of node A refers to node B, then A is data-dependent on B. For instance, in the following code snippet:
```

doc/main/scala/lms/core/backend.md
```diff
@@ -128,25 +114,33 @@ x = 10
...
 the `x = 3` node can be removed without breaking the code semantics. The order of
-graph construction (the fact that user did `x = 3` before `x = 10`) is the base of
-our control dependency analysis.
+graph construction (the fact that the user did `x = 3` before `x = 10`) is the base of our control dependency analysis.
+
+### Effects and EffectSummary
+
+How can we obtain data and control dependency information from the source program? As we mentioned before, data dependency is already captured by the definition of `Node`. However, it is not clear how to collect control dependencies. It turns out that we can track the *effects* of nodes and blocks and then compute control dependencies from these effects. To make our analysis more fine-grained (so that more interesting code reordering and optimization can be introduced), the first step is to realize that there are different kinds of effects. The effect of printing is clearly not so related to the effect of variable reading. Thus there should be no scheduling enforcement between them.
+In LMS IR, we track *read* and *write* effects. They are categorized in the following cases:
```
Kraks, line 125: I guess it would be good to point out that these "phantom" effect keys (such as STORE/CTRL) can be user-defined too.

doc/main/scala/lms/core/backend.md
```diff
@@ -160,35 +154,31 @@ call of the function to print `hello world`. However, the print statement doesn'
...
-`inputs` to the `Block` (`in: List[Sym]`), the `results` of the `Block` (`res: Exp`),
-the `input-effect` of the `Block` (`ein: Sym`), and the `effects` of the `Block`
-(`eff: EffectSummary`).
+1. Read-After-Write (RAW): there should be a hard dependency from R to W,
+   since the correctness of the R depends on the W to be scheduled)
+2. Write-After-Read (WAR): there should be a soft dependency from W to R,
+  since the correctness of the W does not depend on the R to be scheduled, but the order of them matters.
+3. Write-After-Write (WAW): the first idea is to generate a soft dependency,
+  since the second W simply overwrites the first W.
+  However, write to array is more complicated, such as arr(0) = 1; arr(1) = 2,
```
Kraks, line 168:
```suggestion
  However, write to array is more complicated, such as `arr(0) = 1; arr(1) = 2`,
```

doc/main/scala/lms/core/backend.md
```diff
@@ -581,3 +632,69 @@ withBlockScopedEnv(here){
...
+    used ++= valueSyms(d)
+  } else if (d.eff.hasSimpleEffect || d.eff.wkeys.exists(used)) {
+    used += d.n
+    used ++= valueSyms(d)
+  }
+}
+```
+After getting all reachable nodes in `newNodes`, we continue collect used variables by traversing `newNodes` in reverse order. For each node `d`, if the symbol of `d` is in `used`, we add the `valueSym` of `d` into `used` as well. If `d` is not in `used` but has simple effects (e.g. print) or writes to a used symbol, we also add the symbol and `valueSym` of `d` into `used`.
+
+### Recreate the Graph
```
Kraks, line 679:
```suggestion
### Reconstruct the Graph 
```

doc/main/scala/lms/core/backend.md
```diff
@@ -581,3 +632,69 @@ withBlockScopedEnv(here){
...
+    d
+}
+val newBlock = if (fixDeps)
+  g.block.copy(eff = g.block.eff.filter(used))
+else
+  g.block
+
+Graph(newNodes, newBlock, newGlobalDefsCache)
+```
+Lastly, we recreate the graph based on the information we collected earlier. We only consider nodes in `newNodes` that are also in `used`. For each such node, we update `newGlobalDefsCache` to map the symbol of the node to the node itself. We then remove unused effect symbols from the RHS and `EffectSummary` of the node. We do the same thing to the top-most block and return a new `Graph` with the updated information.
```
Kraks, line 700:
```suggestion
Lastly, we reconstruct the graph based on the information we collected earlier. We only consider nodes in `newNodes` that are also in `used`. For each such node, we update `newGlobalDefsCache` to map the symbol of the node to the node itself. We then remove unused effect symbols from the RHS and `EffectSummary` of the node. We do the same thing to the top-most block and return a new `Graph` with the updated information.
```

Kraks (approved):
Thanks for keep updating the doc! They look great to me, some minor suggestions/comments are attached. 

control dependency vs effect dependency terminology: it would be better to be consistent with our draft paper.

Kraks:
> In the file lms_code_walk.md:
> in high level -> at high level?
> 
> In the file traversal.md:
> This functionality is handled by the `Transformer` class, -> This is handled by the `Transformer` class,?
> I am not sure about the phrasing of control dependencies. I thought there were (effect) dependencies based on the email during the POPL submission.
> We define the bound variables of these new symbol -> We define the bound variables of these new symbols ?
> For each node in `g`, get all bound variables that it depends on -> For each node in `g`, get all bound variables it depends on ?
> then schedule the node outside means that the node -> then scheduling the node outside means that the node?

For typos/small changes, you can directly make suggestion in Github: click the blue + button at the line number you want to suggest the change, then click the ± button in the tab of the edit field. It is easier for people to identify what changes.

feiwang3311 (approved):

PR #113, opened 2021-09-26
Expose sym when using reflectEffect

Kraks:
For a downstream customization, I need to expose the symbol name when calling `reflectEffect`. 

An overloaded function `reflectEffect` with the same old interface is also added to keep compatibility.


Kraks:
superseded by https://github.com/TiarkRompf/lms-clean/pull/124

PR #127, opened 2023-02-06
Alternative scheduler with proper graph traversal

Tantalus13A98B5F:
- LMS graph scheduling  
	- `lookupInner`: a mapping from `Sym` to `Node` with metadata, wrapped as `RepNode`  
		- priority value  
		- ~~`scheHere`~~`reachableHot`: required in the outer scope, at least once  
			- will be scheduled there only if available  
			- verb: `reach` vs `reachInner`  
		- ~~`reachHard`~~`reachableHard`: required by a hard dependency path, at least once  
			- unnecessary, will prune further reachability if unavailable  
			- we only care about the soft deps of a node actually scheduled in the outer scope  
			- verb: `reach` vs `reachSoft`  
	- `ScheduleQueue`  
		- maintain a descending priority queue  
		- no repeated enqueue for performance -- metadata update is in-place  
		- `lookup` has the same scope as `inner`, also serving as scheduling boundary  
		- optimization: share the same `lookup` for blocks sharing the same `inner`  
	- testing  
		- pipeline passed  
		- `echo` whole program: ~170s => ~35s  
		- `echo` application only: ~15s => ~5s  
		- `libc_kl_posix` library only: 8947s => ~71s  

To use, mixin `GraphTraversal` in your choice of transformer or codegen.

### DIFF
--- a/src/main/scala/lms/core/codegen.scala
+++ b/src/main/scala/lms/core/codegen.scala
@@ -14,7 +14,7 @@ class Unknown // HACK: Sentinel for typeMap
  * This `GenericCodeGen` class demonstrates a minimal code generator from Traverser.
  * It is not used in production.
  */
-class GenericCodeGen extends Traverser {
+class GenericCodeGen extends Traverser with GraphTraversal {
 
   def emit(s: String) = println(s)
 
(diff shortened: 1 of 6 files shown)


TiarkRompf:
@supunab can you verify that Flare codegen works with Songlin's revised traversal/codegen?

supunab:
Yup, I'll have a look

Kraks:
> Should the standard code generators use this version by default, or should it be opt-in?

I think it looks good to just use this version by default.

supunab:
Flare works as expected as well (ran all tpch queries and verified the output)

supunab (approved):

TiarkRompf:
Merged, thank you!

PR #129, opened 2023-07-18
Add provenance tracking for array cells

CT075:
This PR improves tracking of alias information for nested mutable arrays.

Currently, writes to nested mutable structures are extremely likely to be incorrectly marked as dead code. This is because the current means of effect tracking only tracks one "parent" at a time. That is, in the following snippet:

```
// a : Rep[Array[Array[Int]]]
val a0 = a(0)
a0(0) = 2
return a
```

the write to `a0(0)` is correctly associated with the cell `a0`, but that write is not propagated to the parent array `a`, leaving LMS to conclude that the writes are unobservable and can be optimized out.

This PR seeks to fix this by adding a new parameter to the `Wrap` class, `provenance`, holding the set of all cells from which the wrapped node originated. For example, the value `a0` has provenance `[a]`. Now, when writing to `a0(0)`, we can properly mark the write as also affecting `a`.

Presumably, in the future, this change can be subsumed by a more general mechanism like reachability types, but for now this is a simple mechanism that fixes a bug right now.

### DIFF
--- a/src/main/scala/lms/collection/ArrayOps.scala
+++ b/src/main/scala/lms/collection/ArrayOps.scala
@@ -52,7 +52,6 @@ object ArrayTypeLess {
 }
 
 trait ArrayOps extends PrimitiveOps {
-
   def NewArray[T:Manifest](x: Rep[Int]): Rep[Array[T]] = {
     Wrap[Array[T]](Adapter.g.reflectMutable("NewArray", Unwrap(x)))
   }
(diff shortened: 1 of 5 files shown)


namin:
@Kraks -- any thoughts on this?

TiarkRompf:
This seems useful, but I think I would indeed prefer to see this done in a more general way based on reachability types.

